PR #9, opened 2021-02-08
Add the first Date Import wizard step

tavy88:
### Context

Add the first import date step

This PR introduces the following:

- sets the foundation for the wizard steps were the same
active model object can be shared across multiple questions
- add the govuk form builder
- add the first import date form
- add object validations
- persist the valid import dates on the session using %d/%m/%Y
format
- add Rubocop config

#### Brief

Since we don't persist any of the answers to the questions in the DB, we can't use ActiveRecord. However, in Rails, forms will need an object passed so that validations can be handled properly.

One solution which I have gone with is to use ActiveModel, which leverages all the benefits of ActiveRecord, but without the need of storing any of the data passed through the form. In our case, we will store the data (when valid) on a session.

#### Challenge

The challenge is that we need to carry data from one question to another, through the session, while still using the Rails convention (forms, form objects, validations, etc).

#### Proposed Solution

Use a wizard-like approach, where users go from question to question, step by step. Each step is unique in terms of collected data. 

Resources Strcuture

**Models**

- created a `wizard/steps` namespace where all the active models that map to individual questions will live under
- created a base active model class that sets access to a session and attributes that are being passed from the form to any object
- all the child models that will inherit from the base model just need to worry about the validations of different attributes that are specific to the form they represent and implement the session storage logic under the `save!` method (just like ActiveRecord).

Scalability:

- Let us consider the following set of questions **Q1, Q2, Q3,.., Q7**
- Each question will have its own set of attributes on its corresponding form. E.g
   - Q1: when the answer is valid, it stores the **import_date** value
   - Q2: when the answer is valid, it stores the **country origin** value, etc.
- The validations from Q1 must not be called on Q2 and vice-versa. Each question is unique in its own way. Each question provides unique information and does not overlap with any other question.
- The active model objects must only share the session between them but not the validations or persistence logic. The persistence logic is unique to each question in part
- Clean implementation as each question is decoupled one from another.

**Controllers**

- Controllers that belong to this step by step wizard have also been put under `wizard/steps` namespace to isolate them from the rest of the application controllers. 
- Having separate controllers allows us to use the CRUD actions (index and create) which is the Rails convention
- Ensures simple and dry controllers

#### Conclusion

- Validations are scoped to individual questions that are unique in their own way
- The ActiveModel objects that map to each question are able to share the session between them
- Follows Rails convention and controllers use CRUD default actions
- Easy to add another question form, validations, and session persistence logic
- Keeps code dry, clean and simple
- Decoupled logic and validations for the active model objects


app/controllers/wizard/steps/import_dates_controller.rb
```diff
@@ -0,0 +1,32 @@
+module Wizard
+  module Steps
+    class ImportDatesController < BaseController
+      def index
```
willfish, line 4: So effectively what we're doing here is rendering an individual question answer, right? Any reason we don't use `show` ?

app/models/wizard/steps/base.rb
```diff
@@ -0,0 +1,19 @@
+module Wizard
+  module Steps
+    class Base
+      include ActiveModel::Model
+
+      attr_reader :session,
+                  :attributes
+
+      def initialize(session, attributes = {})
```
willfish, line 9: Could we make the session an active model attribute rather than overriding the constructor?

```ruby
question = Wizard::ImportDate.new(session: session, **attributes)
question.import_date => '2020-01-01'
```

I think overriding the constructor is breaking the normal getter/setter api of ActiveModel::Model which assigns in the constructor all of the values to instance variables or something.

Also happy that the answer might be in a later PR since it would probably break some tests/delay things to do that

app/controllers/wizard/steps/import_dates_controller.rb
tavy88: Because it's a static page it makes more sense to be rendered on the index. For show usually rails expects something like: `/resource/:id`. I know you are thinking about the `:commodity` in the url but `duty-calculator/:commodity/import_date` is not really a show. A show would be something like `duty-calculator/:commodity/import_date/:id`

app/models/wizard/steps/base.rb
willfish: Actually, all we need to do is call super passing the attributes

willfish:
```suggestion
      def initialize(session, attributes = {})
        super attributes
```

tavy88: Let's chat to make sure I understand.

app/controllers/wizard/steps/import_dates_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+        @import_date = Wizard::Steps::ImportDate.new(session)
+      end
+
+      def create
+        @import_date = Wizard::Steps::ImportDate.new(session, permitted_params)
+
+        if @import_date.valid?
+          @import_date.save!
```
willfish, line 12: Do we want the primadonna method, here? I think that effectively would propagate an error normally in the situation where the record was invalid and normally be handled by the base controller which would render any error messages

willfish (approved):

app/controllers/wizard/steps/import_dates_controller.rb
tavy88: Tbh Rubocop was crying about it. I'm not very keen on it.

app/models/wizard/steps/base.rb
tavy88: Btw I like the format you proposed `2020-01-01` better for the date. Will change to use that.

tavy88: I've refactored this.

willfish (approved):

PR #10, opened 2021-02-08
Circle CI deploy to dev on push.

mafonso:
* Add CircleCI deploy config
* Bump node versions and fix dependencies
* Fix webpack config


package.json
```diff
@@ -3,14 +3,15 @@
     "node": "12.x"
   },
   "dependencies": {
-    "@rails/webpacker": "^5.2.1",
+    "@rails/webpacker": "^6.0.0-beta.5",
```
tavy88, line 6: We need to be sure this new version won't cause issues or conflicts with the Gov UK frontend lib.

tavy88 (approved):

package.json
mafonso: Reverted

PR #15, opened 2021-02-11
Add next question placeholder page

tavy88:
### Context

If the answer from the import date question is valid,
move to the next one: import destination.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,6 +1,27 @@
-### Context
+### Jira link
```
willfish, line 1: Yay :tada:

willfish (approved):

PR #18, opened 2021-02-15
[HOTT-440] Service routing

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-440

### What?

I have added/removed/altered:

- [x] Make it so that the service prefixes we expect to see coming from the frontend persist when we redirect back to the frontend
- [x] Tidy up the routes to enable helpers which generate the correct service routes
- [x] Downgrade rails for compatibility
- [x] Tidy up service helper

### Why?

I am doing this because:

- We want redirects to work seamlessly when back is clicked and redirect users to the correct location


.env.development
```diff
@@ -1 +1,4 @@
-TRADE_TARIFF_FRONTEND_ORIGIN=development
\ No newline at end of file
+API_SERVICE_BACKEND_URL_OPTIONS={"uk":"http://localhost:3000","xi":"http://localhost:3000" }
+DUTY_CALCULATOR_HOST=http://test.host
```
willfish, line 2:
```suggestion
DUTY_CALCULATOR_HOST=http://dev.host
```

.env.development
```diff
@@ -1 +1,4 @@
...
+SERVICE_DEFAULT=uk
```
tavy88, line 3: Do we still need this?

.env.test
```diff
@@ -0,0 +1,4 @@
+API_SERVICE_BACKEND_URL_OPTIONS={"uk":"http://localhost:3018","xi":"http://localhost:3019" }
+FRONTEND_HOST=http://test.host
+SERVICE_DEFAULT=uk
```
tavy88, line 3: Do we still need this?

tavy88 (approved):
You can remove SERVICE_DEFAULT I think.

PR #19, opened 2021-02-16
Refactor the ImportDate ActiveModel class & update our session management logic

tavy88:
### Jira link

HOTT-441

### What?

I have added/removed/altered:

- [x] Refactor the ImportDate ActiveModel class
- [x] Remove all the keys from the session that follow the current question

### Why?

We needed to be able to prepopulate the date_select field
with the value from the session.

This is needed for when a user comes back to this question.
In addition, we needed an invalid past date to be persisted
on the form so that users can see what they've done wrong.

Also updating the error message.


app/models/wizard/steps/base.rb
```diff
@@ -9,12 +9,26 @@ class Base
       def initialize(session, attributes = {})
         @session = session
 
+        clean_session if attributes.empty?
```
tavy88, line 12: Ensure we only clean the session while on show action. attributes will be non-empty only on #create actions.

app/models/wizard/steps/base.rb
```diff
@@ -9,12 +9,26 @@ class Base
...
+      private
+
+      def clean_session
+        return if keys_to_remove.empty?
+
+        session.delete(*keys_to_remove)
+      end
+
+      def keys_to_remove
+        session.keys.map(&:to_i).uniq.select { |key| key > self.class::STEP_ID }
```
willfish, line 30: Exactly :)

app/models/wizard/steps/import_date.rb
```diff
@@ -1,38 +1,68 @@
...
+
       attribute 'import_date(3i)', :string
       attribute 'import_date(2i)', :string
       attribute 'import_date(1i)', :string
 
       validate :import_date_in_future
 
-      def import_date
-        session[:import_date]
+      STEP_ID = 1
```
willfish, line 14: Do we want this to be a string?

willfish: I don't think it'll get serialised as an integer

willfish: So we'll be mixing types.

app/views/wizard/steps/import_dates/show.html.erb
```diff
@@ -3,7 +3,7 @@
 <main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-three-quarters">
-      <%= form_for @import_date, builder: GOVUKDesignSystemFormBuilder::FormBuilder, url: import_date_path do |f| %>
+      <%= form_for @import_date, builder: GOVUKDesignSystemFormBuilder::FormBuilder, url: import_date_path, html: { novalidate: true } do |f| %>
```
willfish, line 6: I'm assuming we're skipping validations that are done automatically here? Not sure I understand this change

.gitignore
```diff
@@ -11,6 +11,9 @@
 /log/*
 /tmp/*
 
+# Ignore simplecov files
+/coverage/*
```
willfish, line 15: Would be good at some point to plumb this into the duty calc

app/models/wizard/steps/import_date.rb
tavy88: Uhm sure yeah.

app/views/wizard/steps/import_dates/show.html.erb
tavy88: It's something that GDS wants.

.gitignore
tavy88: Do you mean CircleCi wise? Think Miguel is on it already.

willfish (approved):

PR #20, opened 2021-02-16
[HOTT-430] Integrate commodities via the uktt gem

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-430

### What?

Screenshot from 2021-02-16 15-31-13

I have added/removed/altered:

- [x] Adds the uktt gem to the project after some recent fixes: https://github.com/TransformCore/uktt/pull/1
- [x] Adds model for Commodity which encapsulates the JSON:API response in an openstruct
- [x] Adds builder service which is used to configure between the different services

### Why?

I am doing this because:

- This is needed so that we can add real commodity descriptions
- This will be used as part of fetching data for duty calculator validations later on


app/controllers/wizard/steps/base_controller.rb
```diff
@@ -2,6 +2,18 @@ module Wizard
   module Steps
     class BaseController < ::ApplicationController
       default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
+
+      before_action :assign_commodity
```
tavy88, line 6: Would `fetch_commodity` be a better name here?

app/controllers/wizard/steps/base_controller.rb
```diff
@@ -2,6 +2,18 @@ module Wizard
...
+
+      def assign_commodity
+        @commodity = Commodity.new(code: commodity_code)
```
tavy88, line 9: How about if you are coming from the XI service?

spec/rails_helper.rb
```diff
@@ -1,6 +1,5 @@
 ENV['RAILS_ENV'] ||= 'test'
 
-require 'spec_helper'
```
tavy88, line 3: Why are you removing the `spec_helper` here?

spec/spec_helper.rb
```diff
@@ -1,11 +0,0 @@
...
-  config.expect_with :rspec do |expectations|
-    expectations.include_chain_clauses_in_custom_matcher_descriptions = true
-  end
-
-  config.mock_with :rspec do |mocks|
-    mocks.verify_partial_doubles = true
-  end
-
-  config.shared_context_metadata_behavior = :apply_to_host_groups
-end
```
tavy88, line 11: I don't think we should remove this really. We might need these settings in the future. It's something that Rails will generate by default.

app/models/commodity.rb
```diff
@@ -0,0 +1,29 @@
...
+               end
+  end
+
+  private
+
+  attr_reader :response, :service
+
+  def options
+    @options = OptionBuilder.new(service).call
+      .merge(commodity_id: code)
```
tavy88, line 27: For consistency, I think this should be `commodity_code` here.

app/models/commodity.rb
```diff
@@ -0,0 +1,29 @@
...
+  def description
+    fetch.data.attributes.description
+  end
+
+  def fetch
+    @fetch ||= begin
+                 commodity = Uktt::Commodity.new(options)
+                 commodity.retrieve
+                 commodity.response
+               end
```
tavy88, line 18: Because this method actually returns a response, I would suggest naming this method `response` instead. I think it reads better.

app/views/wizard/steps/import_dates/show.html.erb
```diff
@@ -9,7 +9,7 @@
         <%= f.govuk_date_field :import_date, legend: { text: 'When will the goods be imported?', size: 'xl', tag: 'h1' }, hint: { text: 'As duties and quotas change over time, it may be important to enter the proposed import date. For example, 27  3  2021' } %>
         <%= f.govuk_submit %>
       <% end %>
-      <%= render 'shared/commodity_details' %>
+      <%= render 'shared/commodity_details', locals: { commodity_code: @commodity.code, commodity_description: @commodity.description } %>
```
tavy88, line 12: I think `@commodity` should be accessible inside that partial? Not sure if you need to explicitly pass it, as this variable will be available at the controller level.

.env.development
```diff
@@ -1,3 +1,3 @@
-API_SERVICE_BACKEND_URL_OPTIONS={"uk":"http://localhost:3000","xi":"http://localhost:3000" }
+API_SERVICE_BACKEND_URL_OPTIONS={"uk":"https://dev.trade-tariff.service.gov.uk","xi":"https://dev.trade-tariff.service.gov.uk/xi"}
```
tavy88, line 1: This can be 

```ruby
API_SERVICE_BACKEND_URL_OPTIONS={uk:"https://dev.trade-tariff.service.gov.uk", xi:"https://dev.trade-tariff.service.gov.uk/xi"}
```
Then you don't need to symbolize this later in the code where you use it.

config/application.rb
```diff
@@ -5,15 +5,20 @@
...
   class Application < Rails::Application
     config.load_defaults 5.2
     config.middleware.use Rack::Deflater
+    config.time_zone                 = 'London'
     config.exceptions_app            = routes
     config.trade_tariff_frontend_url = ENV['TRADE_TARIFF_FRONTEND_URL']
     config.duty_calculator_host      = ENV.fetch('DUTY_CALCULATOR_HOST', 'http://localhost')
+    config.api_options               = JSON.parse(ENV['API_SERVICE_BACKEND_URL_OPTIONS']).symbolize_keys
+    # config.uk_api_options            = OptionBuilder.new(:uk).call
+    # config.xi_api_options            = OptionBuilder.new(:xi).call
```
tavy88, line 22: Can we remove these?

.env.development
willfish:
```
json = '{uk:"https://dev.trade-tariff.service.gov.uk", xi:"https://dev.trade-tariff.service.gov.uk/xi"}'
=> "{uk:\"https://dev.trade-tariff.service.gov.uk\", xi:\"https://dev.trade-tariff.service.gov.uk/xi\"}"
[8] pry(main)> JSON.parse(json)
JSON::ParserError: 783: unexpected token at '{uk:"https://dev.trade-tariff.service.gov.uk", xi:"https://dev.trade-tariff.service.gov.uk/xi"}'
from /home/william/.asdf/installs/ruby/2.7.0/lib/ruby/2.7.0/json/common.rb:156:in `parse'
```

app/controllers/wizard/steps/base_controller.rb
willfish: We're not doing the fetch yet, though. We lazy load the response for when we need it in, say, a specific view or controller

app/models/commodity.rb
willfish: Yeah, there was something fishy about it to me, too. I wanted it to have a verb as a name rather than a noun to reflect that it was doing something. But I think it's fine for it to feel like a getter.

app/models/commodity.rb
willfish: But the option buillder might need to merge in other ids so we'd be losing flexibility for a small naming difference

willfish: So the Uktt gem expects commodity_id not commodity_code if that wasn't clear

app/views/wizard/steps/import_dates/show.html.erb
willfish: Yeah, I couldn't see obvious idioms on the web beyond the rails docs wanting you to use locals with partials. These partials also have access to params, all the methods in the base controller and all the methods in current controller.

spec/rails_helper.rb
willfish: Isn't needed

willfish: So we return open struct in the api which is convenient but difficult to stub in the tests since we weren't allowing unverifying doubles. I'd literally have to define each level of OpenStruct as an object and fill it with what I needed. My choices were:

1. Move to pure json - lose the convenience of open struct
2. Remove the verifying partial line in the spec - cleaner but we're not even using the spec helper atm
3. Add all of the open struct definitions - painful and verbose and not adding a lot of value
4. Remove the spec helper

app/models/commodity.rb
tavy88: Sure you can use
```ruby
{
      host: api_options[service],
      version: 'v2',
      debug: debug?,
      return_json: false,
      commodity_id: commodity_code
    }
```
Then? Also, don't worry about other ids we'll tackle them as they come along. Not sure if in reality, we might need anyhting else but this tbh.

.env.development
tavy88: Ah interesting. Fair!

app/helpers/service_helper.rb
```diff
@@ -11,6 +11,10 @@ def tools_url
     service_url_for('/tools')
   end
 
+  def commodity_url(commodity_code)
+    service_url_for("/commodities/#{commodity_code}")
```
tavy88, line 15: Nice!

app/models/commodity.rb
willfish: Yeah, probably you're right about chapters, sections, etc, not being needed

app/helpers/service_helper.rb
```diff
@@ -24,6 +28,8 @@ def trade_tariff_frontend_url
   end
 
   def service_choice
-    params[:service_choice] == 'uk' ? '' : params[:service_choice]
+    return '' if params[:service_choice] == 'uk' || params[:service_choice].nil?
```
willfish, line 31:
```suggestion
    return '' if params[:service_choice] == 'uk'
```

tavy88 (approved):
Looks great!

PR #31, opened 2021-02-22
Adding UserSession wrapper around the session

tavy88:
### Jira link

HOTT-452

### What?

I have added/removed/altered:

- [x] Use more readable ids for GB and NI
- [x] Add UserSession wrapper around the session

### Why?

- Because we need to do some intensive logic around the country of import, it's easier to store the country codes on the session rather than incremental ids. We could not use geographical area ids because we don't have the entries for these 2 options.
- The session should not be accessed directly within the models.

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -1,0 +1,41 @@
+class UserSession
+  attr_reader :session
+
+  def initialize(session)
+    @session = session
+  end
+
+  def remove_keys_after(key)
+    keys_to_remove = session.keys.map(&:to_i).uniq.select { |k| k > key }
+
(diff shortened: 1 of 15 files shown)


app/models/user_session.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def remove_keys_after(key)
+    keys_to_remove = session.keys.map(&:to_i).uniq.select { |k| k > key }
+
+    return if keys_to_remove.empty?
+
+    session.delete(*keys_to_remove)
+  end
+
+  def import_date
```
willfish, line 16: This is so much better

willfish (approved):

PR #32, opened 2021-02-23
This PR adds the following:

tavy88:
### Jira link

HOTT-452

### What?

- use @step as ActiveModel object for any of the
wizard steps: import_date, import_destination, etc. It
makes more sense this way
- add next_step_path and previous_step_path methods on every
ActiveModel class, as we want to tackle the pathfinder logic
at each step in part, rather than have one big class with
all the rules
- add the first duty calculator version for NI to GB route
- add the duty page for NI to GB route
- use XI as area code for NI for consistency across the platform

### DIFF
--- a/app/models/wizard/steps/base.rb
+++ b/app/models/wizard/steps/base.rb
@@ -22,7 +22,7 @@ module Wizard
       protected
 
       def next_step_path(*)
-        raise('Must implement this method')
+        raise NotImplementedError
       end
 
       def previous_step_path(*)
--- a/app/models/wizard/steps/base.rb
+++ b/app/models/wizard/steps/base.rb
@@ -3,6 +3,7 @@ module Wizard
     class Base
       include ActiveModel::Model
       include ActiveModel::Attributes
+      include Rails.application.routes.url_helpers
 
       attr_reader :user_session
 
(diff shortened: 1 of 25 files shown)


app/models/wizard/steps/base.rb
```diff
@@ -18,6 +19,16 @@ def persisted?
         false
       end
 
+      protected
+
+      def next_step_path(*)
+        raise('Must implement this method')
```
willfish, line 25: Ruby has a specific thing for non-compliance with interfaces: https://ruby-doc.org/core-2.5.0/NotImplementedError.html

willfish:
```suggestion
        raise NotImplementedError
```

app/models/wizard/steps/base.rb
```diff
@@ -18,6 +19,16 @@ def persisted?
...
+      end
+
+      def previous_step_path(*)
+        raise('Must implement this method')
```
willfish, line 29:
```suggestion
        raise NotImplementedError
```

app/views/wizard/steps/duties/_no_duty.html.erb
```diff
@@ -0,0 +1,4 @@
+<% if @user_session.ni_to_gb_route? %>
```
willfish, line 1: Feels weird this being a view on a step controller :confused:

willfish (approved):
Freaking awesome man!

app/models/wizard/steps/base.rb
tavy88: 👍

PR #33, opened 2021-02-23
Implement the EU to NI route

tavy88:
### Jira link

HOTT-470

### What?

Implement the EU to NI route

### DIFF
--- a/app/models/country.rb
+++ b/app/models/country.rb
@@ -1,6 +1,8 @@
 class Country
   attr_reader :code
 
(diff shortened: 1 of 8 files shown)


app/models/country.rb
```diff
@@ -1,6 +1,8 @@
 class Country
   attr_reader :code
 
+  EU_MEMBERS = %w[AT BE BG CY CZ DE DK EE ES EU FI FR GR HR HU IE IT LT LU LV MT NL PL PT RO SE SI SK XI].freeze
```
tavy88, line 4: This will be replaced by the API endpoint call when that is ready.

willfish (approved):

PR #35, opened 2021-02-25
[HOTT-476] JSON Api models

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-476

### What?

I have added/removed/altered:

- [x] Pull out geographical area eu members
- [x] Integrate with jsonapi responses

### Why?

I am doing this because:

- We need an easy to use interface for our resources


app/models/api/base.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    def self.has_one(relation, klass)
+      define_method(relation) do
+        klass.new(attributes[relation.to_s])
+      end
+    end
+
+    def self.resource_key(key)
+      define_singleton_method(:resource_key) do
```
tavy88, line 27:
```suggestion
      define_singleton_method(:resource_key) { key }
```

tavy88 (approved):
:shipit:

PR #37, opened 2021-02-25
Pin buildpack

mafonso:
### Jira link

n/a

### What?

I have added/removed/altered:

- [ ] Pining the CloudFoundry build pack version

### Why?

I am doing this because:

- We want the build pack ruby and bundler versions to match the ones were pinned in the repo.


mafonso:
Moved into #36

PR #40, opened 2021-02-26
This PR adds the following:

tavy88:
### Jira link

HOTT-479

### What?

- The new trader scheme page
- Renames some of our controllers

### Why?

Some controller names made sense to use the singular form e.g ImportDateController.


mafonso (approved):
Reviewed on a call with @tavy88 @jcavell and @mafonso

PR #41, opened 2021-03-01
Add additional validations for customs value page

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- Add additional validations for customs value page
- Treat cases where the values might be negative.
- Set customs value class to inherit from the right parent class.

### Why?

Users should not add negative values on the form.


mafonso (approved):
LGTM

PR #42, opened 2021-03-01
Define which step ids to remove from the session

tavy88:
### Jira link

HOTT-<TODO>

### What?

This change allows devs to define which questions to be
removed from the session while positioned on a particular step.
This way the STEP_ID now becomes purely a unique identifier
for that step.

### Why?

Because some of the questions will be shared by
other routes, we can no longer rely on the position
of that question in the wizard.

The same question could be on position 4 or 8 depending
on which route the user lands on.

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -5,12 +5,8 @@ class UserSession
     @session = session
   end
 
-  def remove_keys_after(key)
-    keys_to_remove = session.keys.map(&:to_i).uniq.select { |k| k > key }
-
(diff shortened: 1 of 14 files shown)


mafonso (approved):
LGTM 👍

cedrickisemadigitalhmrc:
Looks good to me. We don't have a Jira for this so perhaps we don't need an empty jira link?

cedrickisemadigitalhmrc (approved):

jcavell (approved):

PR #43, opened 2021-03-01
Copy change for Northern Ireland option

tavy88:
Just correct the copy on the second option for import destination page.


mafonso (approved):
LGTM

PR #47, opened 2021-03-04
Support meta_attribute dsl

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Encapsulate meta attributes we're interested in behind a DSL

### Why?

I am doing this because:

- We want an easy way to fetch them for the duty calculator and we'll almost certainly be adding more

### DIFF
--- a/app/models/api/base.rb
+++ b/app/models/api/base.rb
@@ -1,7 +1,20 @@
 module Api
   class Base
-    include ActiveModel::Model
-    include ActiveModel::Attributes
+    def self.inherited(child)
+      child.include ActiveModel::Model
+      child.include ActiveModel::Attributes
+
+      child.attribute :meta
+    end
+
+    def self.meta_attribute(*attribute_path)
+      attribute_path = attribute_path.map(&:to_s)
+      method_name = attribute_path.last
+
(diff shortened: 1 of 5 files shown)


app/models/api/commodity.rb
```diff
@@ -5,6 +5,9 @@ class Commodity < Api::Base
     has_many :import_measures, Measure
     has_many :export_measures, Measure
 
+    meta_attribute :duty_calculator, :zero_mfn_duty
+    meta_attribute :duty_calculator, :trade_defence
+
```
tavy88, line 10: Very nice!

app/models/api/base.rb
```diff
@@ -1,7 +1,20 @@
...
+    end
+
+    def self.meta_attribute(*attribute_path)
+      attribute_path = attribute_path.map(&:to_s)
+      method_name = attribute_path.last
+
+      define_method(method_name) do
+        meta.dig(*attribute_path)
+      end
+    end
```
tavy88, line 17: Could we have some unit tests here?

tavy88 (approved):

PR #51, opened 2021-03-05
Connect the Trader Scheme step to the user flow

tavy88:
### Jira link

HOTT-495

### What?

This PR introduces the following:
- adds the next step path logic for the Country of Origin
question depending on whether there are:
- trade defence measures in place
- zero mfn duty

### DIFF
--- a/app/controllers/wizard/steps/base_controller.rb
+++ b/app/controllers/wizard/steps/base_controller.rb
@@ -12,13 +12,11 @@ module Wizard
         )
       end
 
-      def xi_commodity_for(geographical_area_id)
-        query = { 'filter[geographical_area_id]' => geographical_area_id }
-
+      def filtered_commodity(service_source:, filter:)
         Api::Commodity.build(
-          :xi,
+          service_source,
           commodity_code,
-          query,
+          filter,
         )
       end
 
--- a/app/controllers/wizard/steps/base_controller.rb
+++ b/app/controllers/wizard/steps/base_controller.rb
@@ -3,12 +3,20 @@ module Wizard
     class BaseController < ::ApplicationController
       default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
-      before_action :assign_commodity
+      helper_method :commodity
 
-      def assign_commodity
-        @commodity = Api::Commodity.build(
+      def commodity
+        @commodity ||= Api::Commodity.build(
           service_choice,
           commodity_code,
+        )
+      end
+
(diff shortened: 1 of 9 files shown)


app/controllers/wizard/steps/base_controller.rb
```diff
@@ -3,15 +3,23 @@ module Steps
     class BaseController < ::ApplicationController
       default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
-      before_action :assign_commodity
+      helper_method :commodity
```
willfish, line 6: Nice

app/controllers/wizard/steps/base_controller.rb
```diff
@@ -3,15 +3,23 @@ module Steps
...
-      def assign_commodity
-        @commodity = Api::Commodity.build(
+      def commodity
+        @commodity ||= Api::Commodity.build(
           service_choice,
           commodity_code,
         )
       end
 
+      def xi_commodity_for(geographical_area_id)
```
willfish, line 15: Isn't this only used for the origin step? I think it's cleaner to move it out to the one place it's used?

app/controllers/wizard/steps/country_of_origin_controller.rb
```diff
@@ -18,6 +18,23 @@ def permitted_params
           :country_of_origin,
         )
       end
+
+      def opts
+        return opts_for_ni_to_gb_route if answer == 'GB' && user_session.import_destination == 'XI'
```
willfish, line 23: Makes sense to me. I'm not a fan of the step having a different constructor from the base step but I guess we have to pick a poison somewhere.

willfish:
Feels okay to me. Seems like we've got two choices:

1. Try and come up with a better abstraction for how we make step decisions for virtual steps
2. Move on with this as a reasonable effort that's cleaner than putting everything into the controller

willfish:
I'm in favour of 2 for the MVP I think.

app/controllers/wizard/steps/base_controller.rb
tavy88: I think it will be shared by other routes as well.

app/controllers/wizard/steps/country_of_origin_controller.rb
tavy88: Yeah I think it's ok.

willfish (approved):

willfish (approved):

PR #56, opened 2021-03-10
Add GDS compliant error message

tavy88:
### Jira link

HOTT-<TODO>

### What?

Add GDS compliant error message

### Why?

We need to be GDS compliant


mafonso (approved):
LGTM

PR #57, opened 2021-03-10
[HOTT-410] Measure amounts

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-410

### What?

I have added/removed/altered:

- [x] Adds dynamic attribute implementation which enables the resource we pass to the form builder as well as the validations to be dynamic
- [x] Adds interface for inputting the measure amounts for given units

### Why?

I am doing this because:

- This is needed to generate the components for inputting measure amounts

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -80,7 +80,7 @@ class UserSession
   end
 
   def measure_amount
-    session[Wizard::Steps::MeasureAmount.id]
+    session[Wizard::Steps::MeasureAmount.id] || {}
   end
 
   def monetary_value
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -75,6 +75,14 @@ class UserSession
     }
   end
 
+  def measure_amount=(values)
+    session[Wizard::Steps::MeasureAmount.id] = values
+  end
+
+  def measure_amount
+    session[Wizard::Steps::MeasureAmount.id]
+  end
+
   def monetary_value
     session[Wizard::Steps::CustomsValue.id].try(:[], 'monetary_value')
   end
--- a/app/models/wizard/steps/base.rb
+++ b/app/models/wizard/steps/base.rb
@@ -73,23 +73,13 @@ module Wizard
 
         attributes.each do |attribute|
           define_singleton_method attribute do
-            @dynamic_attributes_instance.public_send(attribute) || @user_session.dig(id, attribute)
+            @dynamic_attributes_instance.public_send(attribute) || @user_session.measure_amount[attribute]
           end
         end
       end
 
       def self.id
         name.split('::').last.underscore
-      end
-
-      protected
-
-      def next_step_path(*)
-        raise NotImplementedError
-      end
-
-      def previous_step_path(*)
-        raise NotImplementedError
       end
 
       def self.dynamic_attributes(dynamic_attribute)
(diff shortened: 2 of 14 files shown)


app/models/wizard/steps/base.rb
```diff
@@ -1,24 +1,83 @@
...
+          @dynamic_attributes_instance.errors
+        end
+
+        attributes.each do |attribute|
+          define_singleton_method attribute do
+            @dynamic_attributes_instance.public_send(attribute) || @user_session.dig(id, attribute)
+          end
+        end
+      end
+
```
tavy88, line 80: I know this is still WIP but I think we should avoid touching the parent class. We only have one ActiveModel class that requires dynamic attributes and I think that should be scoped to that class only.

We will have access to all the attributes that are being submitted from the form via `attributes`. 

https://stackoverflow.com/questions/16571392/activemodel-dynamic-attributes might help?

willfish: Way too late for that :)

willfish: Re validations they're what Matt wants after chatting with him

willfish:
So I'm leaving the feature spec until we plumb together the other parts of this path for now (same as what we've done with the customs value). I guess probably what I suggested around a refactor using a session factory with traits makes more sense than answering every question over and over as we go through each of the step features - one for much later after the initial MVP I guess.

app/models/wizard/steps/base.rb
tavy88: Sure not too fuss about validations just wanted to make our lives easier as much as possible.

app/models/wizard/steps/measure_amount.rb
```diff
@@ -0,0 +1,74 @@
+module Wizard
+  module Steps
+    class MeasureAmount
+      STEPS_TO_REMOVE_FROM_SESSION = %w[].freeze
+
+      include ActiveModel::Model
```
tavy88, line 6: Do we need this here anymore? This is no longer an ActiveModel class now.

spec/models/wizard/steps/measure_amount_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+        },
+      },
+    ).permit!
+  end
+
+  let(:measure_amount) { { 'dtn' => 500.42, 'hlt' => 204.64 } }
+
+  describe 'STEPS_TO_REMOVE_FROM_SESSION' do
+    it 'returns the correct list of steps' do
+      expect(described_class::STEPS_TO_REMOVE_FROM_SESSION).to eq(%w[])
```
tavy88, line 51: to be empty?

tavy88 (approved):
Very nice!

app/models/user_session.rb
```diff
@@ -75,6 +75,14 @@ def customs_value=(values)
...
   end
 
+  def measure_amount=(values)
+    session[Wizard::Steps::MeasureAmount.id] = values
+  end
+
+  def measure_amount
+    session[Wizard::Steps::MeasureAmount.id] || {}
+  end
+
```
tavy88, line 85: Unit test for these please?

app/models/wizard/steps/measure_amount.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+
+      def previous_step_path(service_choice:, commodity_code:)
+        customs_value_path(service_choice: service_choice, commodity_code: commodity_code)
+      end
+
+      def self.id
+        name.split('::').last.underscore
+      end
+
```
tavy88, line 65:
```suggestion
def persisted?
        false
      end
```

Think you need this too otherwise when you try to save I think it will break?

tavy88 (approved):
:shipit:

PR #59, opened 2021-03-11
Use the correct error message

tavy88:
### Jira link

HOTT-<TODO>

### What?

Use the correct error message

### Why?

Original message was wrong as there are 4 options.


willfish (approved):

mafonso (approved):
LGTM

PR #62, opened 2021-03-15
Connect Measure Amount page to the user flow

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [ ] Connect Measure Amount page to the user flow
- [ ] Also add the placeholder page for confirmation

### DIFF
--- a/app/models/wizard/steps/measure_amount.rb
+++ b/app/models/wizard/steps/measure_amount.rb
@@ -48,11 +48,11 @@ module Wizard
       end
 
       def save
-        user_session.measure_amount = answers
+        user_session.measure_amount = answers if answers.present?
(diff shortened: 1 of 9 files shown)


app/models/wizard/steps/measure_amount.rb
```diff
@@ -48,11 +48,11 @@ def initialize(user_session, attributes)
       end
 
       def save
-        user_session.measure_amount = answers
+        user_session.measure_amount = answers if answers.present?
```
tavy88, line 51: We don't want to store a key on the session if there are no answers available.

willfish (approved):

PR #63, opened 2021-03-15
Adds exchange rate model

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-463

### What?

I have added/removed/altered:

- [x] Updates to the latest version of Uktt and uses new api
- [x] Adds exchange rate model

### Why?

I am doing this because:

- We need this for the confirm page


spec/models/api/commodity_spec.rb
```diff
@@ -10,12 +10,6 @@
     allow(Uktt::Commodity).to receive(:new).and_call_original
   end
 
-  describe '.resource_key' do
```
willfish, line 13: This is no longer required since we pass positional config to the retrieve method and the base resource and id is the name of the resource id

app/models/api/commodity.rb
```diff
@@ -25,9 +23,7 @@ class Commodity < Api::Base
                :section,
                :chapter,
                :heading,
-               :ancestors,
-               :import_measures,
```
willfish, line 29: Setters/getters are defined in has_many so removed these

app/models/api/base.rb
```diff
@@ -36,32 +36,29 @@ def self.has_one(relation, klass)
       end
     end
 
-    def self.resource_key(key)
-      define_singleton_method(:resource_key) { key }
-    end
-
     def self.build(service, id, query = {})
-      options = OptionBuilder.new(service).call
+      client = http_client_for(service)
```
tavy88, line 40: Nice!

tavy88 (approved):

PR #64, opened 2021-03-16
Connect the confirmation page to the user flow

tavy88:
### Jira link

HOTT-411

### What?

I have added/removed/altered:

- Connect the confirmation page to the user flow
- Also refactor the keys on the session to now be nested under answers


willfish (approved):
Awesome

PR #65, opened 2021-03-18
[HOTT-525] Fixes asset prefix path in webpack.yml

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-525

### What?

I have added/removed/altered:

- [x] Updates webpacker configuration to handle a path prefix
- [x] Make mapped routes reflect this change

### Why?

I am doing this because:

- So that the frontend can load assets properly


mafonso (approved):
Awesome! 
LGTM

PR #90, opened 2021-04-10
Skip measure amount page if no applicable units

tavy88:
### Jira link

HOTT-530

### What?

I have added/removed/altered:

- [x] Skip measure amount page if no applicable units

### Why?

I am doing this because:

- There are cases when a commodity does not have
applicable measure units available. In this case, users
should jump straight to the confirmation page.

### DIFF
--- a/app/models/wizard/steps/confirmation.rb
+++ b/app/models/wizard/steps/confirmation.rb
@@ -8,7 +8,9 @@ module Wizard
       end
 
       def previous_step_path
-        return measure_amount_path if user_session.gb_to_ni_route?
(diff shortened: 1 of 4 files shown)


app/models/wizard/steps/confirmation.rb
```diff
@@ -8,7 +8,9 @@ def next_step_path
       end
 
       def previous_step_path
-        return measure_amount_path if user_session.gb_to_ni_route?
+        return measure_amount_path unless user_session.measure_amount.empty?
```
willfish, line 11: Nice

willfish (approved):

PR #97, opened 2021-04-12
Fix copy when certificate of origin is present

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [ ] Fix copy when a certificate of origin is present

### Why?

I am doing this because:

- Fix copy when a certificate of origin is present

### DIFF
--- a/app/views/wizard/steps/duty/gb_to_ni/_no_duty.html.erb
+++ b/app/views/wizard/steps/duty/gb_to_ni/_no_duty.html.erb
@@ -27,5 +27,5 @@
     <li>You are transporting goods from England, Scotland or Wales to Northern Ireland</li>
     <li>You are able to take advantage of the preferential tariffs provided by the UK / EU Trade and Co-operation Agreement (TCA) and have a valid Certificate of Origin</li>
   </ul>
-  <%= render('wizard/steps/duty/gb_to_ni/shared/note') %>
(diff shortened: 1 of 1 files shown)


app/views/wizard/steps/duty/gb_to_ni/_no_duty.html.erb
```diff
@@ -27,5 +27,5 @@
     <li>You are transporting goods from England, Scotland or Wales to Northern Ireland</li>
     <li>You are able to take advantage of the preferential tariffs provided by the UK / EU Trade and Co-operation Agreement (TCA) and have a valid Certificate of Origin</li>
   </ul>
-  <%= render('wizard/steps/duty/gb_to_ni/shared/note') %>
+  <p class="govuk-body-m">You may be called upon to provide a copy of your Certificate of Origin to avoid paying duties.<p>
```
willfish, line 30: I'm going to guess that when this partial is rendered it implicitly means `certificate == true && uk_trader_scheme == false`?

willfish (approved):

app/views/wizard/steps/duty/gb_to_ni/_no_duty.html.erb
tavy88: Yep

PR #100, opened 2021-04-12
Cache test

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Add additional headers to suppress CDN cache

### Why?

I am doing this because:

- no-cache header was not enough as it turns out.


willfish (approved):
Assuming this plays nice with the DC I'm happy with but confused by it

PR #102, opened 2021-04-13
Adds footnote per measure type id

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-572

### What?

I have added/removed/altered:

- [x] Adds footnote translations for measure types
- [x] Standardise footnotes for every measure type and option.

### Why?

I am doing this because:

- HMRC have asked that we specify a footnote per every measure type such that all the different measure types are made clear to the user


willfish:
Screenshot from 2021-04-13 11-49-18

tavy88 (approved):

PR #110, opened 2021-04-15
Add better copy on the error pages

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Add better copy on the error pages
- [x] Also return 404 for the root path as we don't have a page for that

### Why?

I am doing this because:

- Users should be able to navigate their way around a 500 or similar error and keep being able to use the service

### DIFF
--- a/app/views/errors/internal_server_error.html.erb
+++ b/app/views/errors/internal_server_error.html.erb
@@ -1,8 +1,9 @@
 <main role="main" class="govuk-main-wrapper" id="main-content">
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <h1 class="govuk-heading-xl">We're sorry, but something went wrong.</h1>
-      <p class="govuk-body">If you are the application owner check the logs for more information.</p>
+      <h1 class="govuk-heading-xl">Sorry, there is a problem with the service</h1>
+      <p class="govuk-body">We could not proceed with calculating your duty.</p>
+      <p class="govuk-body">Click here to <%= link_to('start again', import_date_path) %>.</p>
     </div>
   </div>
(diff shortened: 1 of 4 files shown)


app/views/errors/internal_server_error.html.erb
```diff
@@ -1,8 +1,9 @@
 <main role="main" class="govuk-main-wrapper" id="main-content">
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <h1 class="govuk-heading-xl">We're sorry, but something went wrong.</h1>
-      <p class="govuk-body">If you are the application owner check the logs for more information.</p>
+      <h1 class="govuk-heading-xl">Sorry, there is a problem with the service</h1>
+      <p class="govuk-body">We could not proceed with calculating your duty.</p>
+      <p class="govuk-body">Click here to <%= link_to('start again', import_date_path) %>.</p>
```
willfish, line 6: Should we pass the query params to the import_date_path if they're available?

config/routes.rb
```diff
@@ -1,4 +1,6 @@
 Rails.application.routes.draw do
+  root to: proc { [404, {}, ['Not found.']] }
```
willfish, line 2: Should the mimetype of this be html?

willfish (approved):

app/views/errors/internal_server_error.html.erb
tavy88: Yeah, interesting point. They would already have it on the session though?

config/routes.rb
tavy88: Does it matter that much?

PR #124, opened 2021-04-21
Adds waiver option to gb to ni calculation result

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-576

### What?

I have added/removed/altered:

- [x] Adds waiver default option for GB to NI duty calculations
- [x] Rename config/locales/measure_types.en.yml -> config/locales/footnotes.en.yml

### Why?

I am doing this because:

- This is required to indicate that under certain circumstances the duties into NI can be waived

### ROW -> UK duty calculation

Screenshot from 2021-04-21 10-17-56

### GB -> NI duty calculation

Screenshot from 2021-04-21 10-20-13


app/models/duty_calculator.rb
```diff
@@ -57,4 +57,18 @@ def additional_duty_rows
         acc << option_klass.new(measure, user_session, []).option
       end
   end
+
+  def default_options
+    return [waiver_option] if user_session.gb_to_ni_route?
+
+    []
+  end
```
tavy88, line 65: oh, interesting so it's only on GB to NI route.

app/models/duty_calculator.rb
```diff
@@ -57,4 +57,18 @@ def additional_duty_rows
...
+
+  def waiver_option
+    {}.tap do |option|
+      option[:key] = DutyOptions::Waiver.id
+      option[:evaluation] = DutyOptions::Waiver.new(nil, user_session, []).option
+      option[:priority] = DutyOptions::Waiver::PRIORITY
+    end
```
tavy88, line 72: Why not just: 

```ruby
{
   key: DutyOptions::Waiver.id,
  evaluation: DutyOptions::Waiver.new(nil, user_session, []).option,
 ...
}
```

Any reason why we need to use tap?

tavy88 (approved):

app/models/duty_calculator.rb
willfish: Yarp

app/models/duty_calculator.rb
willfish: So no good reason beyond I always wanted to return the option.

PR #127, opened 2021-04-22
Adds factory bot and user session factory

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-630

### What?

I have added/removed/altered:

- [x] Adds factory bot gem
- [x] Adds user session factory

### Why?

I am doing this because:

- This part of a wider tidying up of the code base and will enable easier refactoring of specs and code

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -11,7 +11,7 @@ class UserSession
   end
 
   def import_date
-    return unless session['answers'].key?(Wizard::Steps::ImportDate.id)
+    return if session['answers'][Wizard::Steps::ImportDate.id].blank?
 
     Date.parse(session['answers'][Wizard::Steps::ImportDate.id])
   end
--- a/spec/factories/user_session.rb
+++ b/spec/factories/user_session.rb
@@ -1,0 +1,37 @@
+Dir[File.join(__dir__, '../../app/models/wizard/steps/', '*.rb')].sort.each { |file| require_relative file }
+
+FactoryBot.define do
+  factory :user_session do
+    POSSIBLE_ANSWERS = Wizard::Steps.constants.map { |step| "Wizard::Steps::#{step}".constantize.id }
+
+    commodity_code   { nil }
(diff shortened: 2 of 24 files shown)


spec/factories/user_session.rb
```diff
@@ -0,0 +1,37 @@
...
+      import_date           { nil }
+      import_destination    { nil }
+      measure_amount        { nil }
+      planned_processing    { nil }
+      trader_scheme         { nil }
+    end
+
+    initialize_with do
+      answers = POSSIBLE_ANSWERS.each_with_object({}) do |answer_method, acc|
+        acc[answer_method] = public_send(answer_method) if respond_to?(answer_method) && public_send(answer_method)
```
tavy88, line 28: Do you need to call  

```ruby
&& public_send(answer_method)
```
Here?

spec/factories/user_session.rb
```diff
@@ -0,0 +1,37 @@
+Dir[File.join(__dir__, '../../app/models/wizard/steps/', '*.rb')].sort.each { |file| require_relative file }
+
+FactoryBot.define do
+  factory :user_session do
```
tavy88, line 4: How do you feel about adding traits like e.g gb_to_ni and then you prepopulate the answers as well rather than leaving them all nil by default? Just a thought...

willfish: It's exactly what I planned to do as a follow on PR - using human-readable traits to define the session

app/models/user_session.rb
```diff
@@ -11,7 +11,7 @@ def remove_step_ids(ids)
   end
 
   def import_date
-    return unless session['answers'].key?(Wizard::Steps::ImportDate.id)
+    return if session['answers'][Wizard::Steps::ImportDate.id].blank?
```
willfish, line 14: The import date answer can be nil or an empty string. This covers both these variants.

tavy88 (approved):

PR #128, opened 2021-04-23
Update route for import_date page

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Update route for import_date page

### Why?

I am doing this because:

- We want to clearly document through routing which params are mandatory for each route.

### DIFF
--- a/app/controllers/wizard/steps/import_date_controller.rb
+++ b/app/controllers/wizard/steps/import_date_controller.rb
@@ -4,12 +4,7 @@ module Wizard
       def show
         @step = Wizard::Steps::ImportDate.new(user_session)
 
-        user_session.commodity_code = commodity_code
-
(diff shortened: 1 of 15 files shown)


app/controllers/wizard/steps/import_date_controller.rb
```diff
@@ -4,12 +4,7 @@ class ImportDateController < BaseController
       def show
         @step = Wizard::Steps::ImportDate.new(user_session)
 
-        user_session.commodity_code = commodity_code
-
-        if params[:referred_service].present?
-          user_session.referred_service = params[:referred_service]
-          user_session.commodity_source = params[:referred_service]
-        end
+        save_commodity_data
```
tavy88, line 7: Maybe `persist` is a better name?

PR #135, opened 2021-04-27
Update copy for duty page

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-645

### What?

I have added/removed/altered:

- [x] Updates no duty bullet points with coordinating conjunctions
- [x] Change column width

### Why?

I am doing this because:

- Business request

### Change visualised

Screenshot from 2021-04-27 11-55-51


cacquani (changes requested):
The ticket specifies that the _and_ should be in italics, has that been changed verbally and the ticket never been updated?

willfish:
Screenshot from 2021-04-27 12-05-06

cacquani (approved):

PR #138, opened 2021-04-27
Copy changes for measure types footnotes

tavy88:
### Jira link

HOTT-646

### What?

I have added/removed/altered:

- [x] Update measure types footnotes

### Why?

I am doing this because:

- Business request

NOTE: This PR does not address the logic around linking back to the correct commodity page.


willfish (approved):
Painful :(

tavy88:
Yes, it was...

PR #140, opened 2021-04-28
Link back to the correct Trade Tariff platform

tavy88:
### Jira link

HOTT-646

### What?

I have added/removed/altered:

- [x]  In case we have the following measure type ids:
551, 552, 553, 554, allow users to go back to the
correct import section of the TradeTariff
platform.

### Why?

I am doing this because:

- Better user experience

### DIFF
--- a/app/services/duty_options/base.rb
+++ b/app/services/duty_options/base.rb
@@ -11,7 +11,7 @@ module DutyOptions
 
     def option
       {
-        footnote: localised_footnote_for(measure.measure_type.id).html_safe,
+        footnote: localised_footnote.html_safe,
         warning_text: nil,
         values: option_values,
       }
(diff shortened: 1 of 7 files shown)


willfish (approved):

app/services/duty_options/base.rb
```diff
@@ -1,6 +1,9 @@
 module DutyOptions
   class Base
     include ActionView::Helpers::NumberHelper
+    include ServiceHelper
+
+    COUNTERVAILING_ANTI_DUMPING_MEASURE_TYPE_IDS = %w[551 552 553 554].freeze
```
willfish, line 6: How do you feel about moving this knowledge to the measure type and exposing it behind a method?

```ruby
REMEDIAL_IDS = %w[551 552 553 554].freeze

def remedial?
  id.in?(REMEDIAL_IDS)
end

# In DutyOptions::Base

def localized_footnote
  return I18n.t("measure_type_footnotes.#{measure_type_id}", link: link) if measure.measure_type.remedial?

  I18n.t("measure_type_footnotes.#{measure.measure_type.id}") 
end
```

willfish: You've always got the measure in scope of a duty option (since they are highly related)

tavy88: yep sounds good!

app/services/duty_options/base.rb
```diff
@@ -99,5 +100,19 @@ def additional_duty_rows
     def additional_duty_values
       additional_duty_options.map { |additional_duty| additional_duty[:value] }
     end
+
+    def localised_footnote_for(measure_type_id)
+      return I18n.t("measure_type_footnotes.#{measure_type_id}") unless measure.measure_type.remedial?
```
willfish, line 105:
```suggestion
      return I18n.t("measure_type_footnotes.#{measure.measure_type.id}") unless measure.measure_type.remedial?
```

willfish (approved):

app/services/duty_options/base.rb
tavy88: 👍

PR #143, opened 2021-05-10
Filter applicable measures

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-662

### What?

I have added/removed/altered:

- [x] Filter in applicable measures:
   - Measures should be included in a calculation if
     - They have no additional code
     - They have an additional code which matches one of our additional code answers
   - Measures should **NOT** be included in a calculation if
     - They have an additional code which does not match one of our additional code answers
- [x] Stop hiding measures that apply to companies

### Why?

I am doing this because:

- We now know the answers for the commodity-extending codes that the user is interested in
- We now no longer need to hide the company-specific measures since we can just filter down to the ones that apply


app/models/duty_calculator.rb
```diff
@@ -71,4 +70,21 @@ def waiver_option
...
+  def applicable_measures
+    @applicable_measures ||= no_additional_code_measures + additional_code_measures
+  end
+
+  def additional_code_measures
+    commodity.import_measures.select do |measure|
+      additional_code = measure.additional_code
+      code_answer = user_session.additional_code[measure.measure_type.id]
+
+      additional_code.present? && code_answer == additional_code.code
```
tavy88, line 83: Nice!

tavy88 (approved):

PR #149, opened 2021-05-18
Build the commodity in one place

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Code Refactor: Add the code that builds a commodity in
one place only and include that helper instead in the
steps base controller.

### Why?

I am doing this because:

- A commodity should only be built in one place


willfish (approved):
Woohoo

PR #151, opened 2021-05-24
HOTT-709: Cosmetic fix to "Trade Remedies apply" page

cacquani:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-709

### What?

I have added/removed/altered:

- [ ] Remove the three-quarters row/column that was further reducing the area occupied by the content on the page

### Why?

The ticket asks to remove the two-thirds section and leave the three-quarters, however the two-thirds section is part of the layout and removing that would have required one of two options:

1. Change the layout for all pages so that they would all be changed from using a two-thirds width to three-quarters
2. Add a new layout file for this page specifically, with no guarantee that it would be used for other content

After a chat with Matt Lavis, the decision was that it was acceptable to remove the three-quarters instead.

### Have you? (optional)

### DIFF
--- a/app/views/pages/trade_remedies.html.erb
+++ b/app/views/pages/trade_remedies.html.erb
@@ -1,13 +1,9 @@
 <%= link_to('Back', country_of_origin_path, class: "govuk-back-link") %>
 
 <main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-three-quarters">
-      <span class="govuk-caption-xl">Calculate import duties</span>
-      <h1 class="govuk-heading-xl">Duties apply to this import</h1>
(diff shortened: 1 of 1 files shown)


app/views/pages/trade_remedies.html.erb
```diff
@@ -1,13 +1,9 @@
 <%= link_to('Back', country_of_origin_path, class: "govuk-back-link") %>
 
 <main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
```
tavy88, line 3: This also needs to go I think. Have a look at the layouts/application.html.

cacquani: You're right, it's duplicated.

I've also added a spacer around the "back" link to make it look less cramped (the 'main' was adding some padding)

No spacer:

Screen Shot 2021-05-24 at 10 45 53

With spacer:

Screen Shot 2021-05-24 at 10 45 31

tavy88 (approved):
Nice

PR #158, opened 2021-05-28
HOTT-718: Filter out GB and EU member states

cacquani:
Filter out GB and EU member states from the 'other' country of origin
step.

### Jira link

https://transformuk.atlassian.net/browse/HOTT-718

### What?

I have added/removed/altered:

- [ ] Added a new method to return only non-eu, non-GB countries in the geographical area model
- [ ] Use the new method to fill in the "other countries" select in the country of origin page 

### Why?

I am doing this because it's pointless to let the user select GB or any EU country from the select when those are covered by the first two radio buttons.

### Have you? (optional)

### DIFF
--- a/app/models/wizard/steps/country_of_origin.rb
+++ b/app/models/wizard/steps/country_of_origin.rb
@@ -34,7 +34,9 @@ module Wizard
         user_session.zero_mfn_duty = zero_mfn_duty
       end
 
-      def self.options_for(service)
+      def self.options_for(service, other = false)
+        return Api::GeographicalArea.other_countries(service.to_sym) if other
+
         Api::GeographicalArea.list_countries(service.to_sym)
       end
 
--- a/app/models/wizard/steps/country_of_origin.rb
+++ b/app/models/wizard/steps/country_of_origin.rb
@@ -34,8 +34,8 @@ module Wizard
         user_session.zero_mfn_duty = zero_mfn_duty
       end
 
-      def self.options_for(service, other = false)
-        return Api::GeographicalArea.other_countries(service.to_sym) if other
+      def self.options_for(service, include_eu_members = true)
+        return Api::GeographicalArea.non_eu_countries(service.to_sym) unless include_eu_members
 
         Api::GeographicalArea.list_countries(service.to_sym)
       end
--- a/spec/models/api/geographical_area_spec.rb
+++ b/spec/models/api/geographical_area_spec.rb
@@ -85,6 +85,29 @@ RSpec.describe Api::GeographicalArea do
     end
   end
 
+  describe '.other_countries' do
+    let(:expected_members) do
+      %w[AT BE BG CY CZ DE DK EE ES EU FI FR GR HR HU IE IT LT LU LV MT NL PL PT RO SE SI SK]
+    end
+    let(:members) { described_class.other_countries.map(&:id) }
+
+    it 'returns non-eu results' do
+      expect(members.length).to be_positive
+    end
+
+    it 'returns none of the member countries of the EU' do
+      expect(members).not_to include(expected_members)
+    end
+
+    it 'doesn\'t return UK' do
+      expect(members).not_to include('UK')
+    end
+
+    it 'doesn\'t return XI' do
+      expect(members).not_to include('XI')
+    end
+  end
+
   describe '.find' do
     it 'returns the country found by id' do
       country = described_class.find(id)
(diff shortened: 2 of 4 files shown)


spec/models/api/geographical_area_spec.rb
```diff
@@ -85,6 +85,29 @@
...
+    it 'returns none of the member countries of the EU' do
+      expect(members).not_to include(expected_members)
+    end
+
+    it 'doesn\'t return UK' do
+      expect(members).not_to include('UK')
+    end
+
+    it 'doesn\'t return XI' do
+      expect(members).not_to include('XI')
```
willfish, line 107: I don't see that this is a feature of this filtering behaviour in the duty calculator?

willfish: Probs can remove this test?

app/models/wizard/steps/country_of_origin.rb
```diff
@@ -34,7 +34,9 @@ def save
         user_session.zero_mfn_duty = zero_mfn_duty
       end
 
-      def self.options_for(service)
+      def self.options_for(service, other = false)
```
tavy88, line 37: I wonder if we could use a better naming here for `other`?

spec/models/api/geographical_area_spec.rb
tavy88: Because XI would never be returned as country on the :xi service right?

willfish: Yarp. It's not in our list of expected eu members

willfish: So we're saying we expect everyone in the list of eu members plus the UK (`GB`) to be rejected. The backend separately excludes `XI` and is independently tested in the backend unit tests for the geographical area service.

So we want one eu member geographical area id to be missing (e.g. `RO`) and one non-eu member geographical area to to be included (e.g. `NI`).

I'm assuming we're relying on the automatically loaded fixtures so we can just pick one from that list? So I'd just pick one of the other countries from that file.

spec/models/api/geographical_area_spec.rb
```diff
@@ -85,6 +85,29 @@
     end
   end
 
+  describe '.other_countries' do
+    let(:expected_members) do
```
willfish, line 89:
```suggestion
    let(:rejected_members) do
```

spec/models/api/geographical_area_spec.rb
```diff
@@ -85,6 +85,29 @@
...
+      %w[AT BE BG CY CZ DE DK EE ES EU FI FR GR HR HU IE IT LT LU LV MT NL PL PT RO SE SI SK]
+    end
+    let(:members) { described_class.other_countries.map(&:id) }
+
+    it 'returns non-eu results' do
+      expect(members.length).to be_positive
+    end
+
+    it 'returns none of the member countries of the EU' do
+      expect(members).not_to include(expected_members)
```
willfish, line 99:
```suggestion
      expect(members).not_to include(rejected_geographical_area_ids)
```

app/models/wizard/steps/country_of_origin.rb
willfish, line 37:
```suggestion
      def self.options_for(service, include_eu_members = true)
```

spec/models/api/geographical_area_spec.rb
cacquani: The "don't include any EU country" is covered by the "returns none of the member countries of the EU" test, I'll add some specific non-EU country

spec/models/api/geographical_area_spec.rb
```diff
@@ -85,6 +85,29 @@
...
+    end
+
+    it 'doesn\'t return UK' do
+      expect(members).not_to include('UK')
```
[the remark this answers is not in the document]
cacquani: True, I changed it in the model but forgot to do it in the specs.

spec/models/api/geographical_area_spec.rb
```diff
@@ -85,6 +85,29 @@
...
   end
 
+  describe '.non_eu_countries' do
+    let(:eu_member_states_ids) do
+      %w[AT BE BG CY CZ DE DK EE ES EU FI FR GR HR HU IE IT LT LU LV MT NL PL PT RO SE SI SK]
+    end
+    let(:members) { described_class.non_eu_countries.map(&:id) }
+
+    it 'returns non-eu results' do
+      expect(members.length).to be_positive
```
willfish, line 95: I think this is implicitly tested on line 98 below so can be removed, too

willfish (approved):

spec/models/api/geographical_area_spec.rb
cacquani: Good point...

PR #162, opened 2021-06-01
Handle expired/empty sessions

tavy88:
If the session expires or is wiped out, then redirect
to /sections on the trade tariff platform.

### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [ ] If the session expires or is wiped out, then redirect
to /sections on the trade tariff platform.

### Why?

I am doing this because:

- Avoid bad user experience (seeing a 500 error) for users when their session expires during their progress.

### DIFF
--- a/app/controllers/wizard/steps/base_controller.rb
+++ b/app/controllers/wizard/steps/base_controller.rb
@@ -2,9 +2,11 @@ module Wizard
   module Steps
     class BaseController < ::ApplicationController
       include CommodityHelper
+      include ServiceHelper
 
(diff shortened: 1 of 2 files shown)


app/controllers/wizard/steps/base_controller.rb
```diff
@@ -58,6 +60,10 @@ def track_session
           referred_service: user_session.referred_service,
         })
       end
+
+      def ensure_session_integrity
+        return redirect_to trade_tariff_url if commodity_code.blank?
```
willfish, line 65: Nice

willfish (approved):

PR #169, opened 2021-06-03
Plumb trade remedies step to RoW to NI journey when there is a trade defence

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] trade remedies become a pseudo step
- [x] plumb trade remedies page to RoW journey when there is a trade defence in place

### Why?

I am doing this because:

- Because the trade remedies page now has logic in place
and we need to know which step to go back to depending
on the journey we are at, it needs to become a step.


willfish (approved):
Definitely a bit painful :sheep: :it:

willfish (approved):

PR #176, opened 2021-06-11
HOTT-744: Put in GTM / GA tagging on the DC

cacquani:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-744

### What?

I have added/removed/altered:

- [ ] Add GTM tagging in the layout
- [ ] Add helper methods to determine the cookies, so that the tagging can be removed if the user opted out

### Why?

I am doing this because we need to track usage on the Duty Calculator

### Have you? (optional)

### DIFF
--- a/app/helpers/cookies_helper.rb
+++ b/app/helpers/cookies_helper.rb
@@ -1,0 +1,11 @@
+module CookiesHelper
+  def policy_cookie
+    cookie = cookies['cookies_policy']
+
+    cookie.present? ? JSON.parse(cookie) : {}
+  end
+
+  def usage_enabled?
+    policy_cookie['usage'] == 'true'
+  end
+end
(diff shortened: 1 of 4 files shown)


willfish (approved):

app/helpers/cookies_helper.rb
```diff
@@ -0,0 +1,11 @@
...
+  def policy_cookie
+    cookie = cookies['cookies_policy']
+
+    cookie.present? ? JSON.parse(cookie) : {}
+  end
+
+  def usage_enabled?
+    policy_cookie['usage'] == 'true'
+  end
+end
```
tavy88, line 11: Do we need this?

cacquani: It's a handy helper that's used in the two partial to decide whether to add the tags to the page or not.
We could just use the test `policy_cookie['usage'] == 'true'` in the partials, but it's just handy to have.

tavy88 (approved):

PR #183, opened 2021-06-14
Move to controller specs

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-742

### What?

Whilst maintaining exactly the same coverage, the point of this PR is to:

- Migrate away from the complex setup of user sessions by reusing a shared example to go through all of the steps; and
- Reduce the overhead to understand all of the dependencies on the session with traits; and 
- To make it faster/easier to develop new steps without the mentioned overhead

Future PRs will consolidate the way we generate api resources using composable traits for our given situation (for example, give me a commodity with applicable vat options, applicable additional codes, etc)

I have added/removed/altered:

- [x] Adds import date controller spec
- [x] Adds country of origin controller spec
- [x] Adds import destination controller spec
- [x] Adds additional code controller spec
- [x] Adds certificate of origin controller spec
- [x] Adds customs value controller spec
- [x] Adds vat controller spec
- [x] Adds final use controller spec
- [x] Adds planned processing controller spec
- [x] Adds trader scheme controller spec
- [x] Adds trade remedies controller spec
- [x] Adds confirmation page controller spec
- [x] Adds measure amount controller spec
- [x] Adds duty controller spec
- [x] Removes duplicating feature specs and related examples

### Why?

I am doing this because:

- The tests have become a bit unwieldy over time.

### AC

- Code coverage hasn't diminished
- Tests are functionally the same just a bit easier to write and understand


app/controllers/wizard/steps/measure_amount_controller.rb
```diff
@@ -31,7 +31,7 @@ def applicable_measure_units
       def measure_amount_answers
         return {} unless params.key?(:wizard_steps_measure_amount)
 
-        params.require(:wizard_steps_measure_amount).permit(*applicable_measure_unit_keys)
+        params.require(:wizard_steps_measure_amount).permit(*applicable_measure_unit_keys).to_h
```
tavy88, line 34: Why is `.to_h` needed here?

willfish: I wanted something I could compare with a hash in the tests. Happy to revert. It's functionally exactly the same from the perspective of the model

spec/controllers/wizard/steps/additional_code_controller_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  let(:session) { build(:user_session, :with_commodity_information) }
+
+  describe 'GET #show' do
+    subject(:response) { get :show, params: { measure_type_id: '105' } }
+
+    it 'assigns the correct step' do
+      response
+      expect(assigns[:step]).to be_a(Wizard::Steps::AdditionalCode)
```
tavy88, line 13: Does `assigns(:step)` not work here?

willfish: Annoying not. I can't put the fetching the response in the before block either as it creates a side affect that breaks other specs

willfish: I tried

```ruby
expect(response; assigns[:step])
```

but that is weird/horrible

tavy88:
Looks good! The only thing that I noticed we are losing with the deletion of the feature specs is testing that the keys from the session are dropped when heading back to a previous question. Any plans around that?

cacquani (approved):
Looks good! 👍

PR #187, opened 2021-06-17
Update VAT option labels on the results page

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Update VAT option labels on the results page

### Why?

I am doing this because:

- So that traders understand better what the VAT options are


.gitignore
```diff
@@ -34,4 +34,6 @@ yarn-debug.log*
 .env.*.local
 *.log
 
+dump.rdb
```
willfish, line 37: :+1:

config/locales/en.yml
```diff
@@ -3,6 +3,11 @@ en:
     formats:
       default: "%d %B %Y"
   number:
+    percentage:
```
willfish, line 6:
```suggestion
    # Used in NumberHelper.number_to_percentage()
    percentage:
```

willfish (approved):

PR #190, opened 2021-06-18
Adds the CommodityContextService

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-770

### What?

I have added/removed/altered:

- [x] Adds a CommodityContextService
- [x] Adds specs for the new service
- [x] Updates the existing CommodityHelper specs to indicate we're now using the CommodityContextService to fetch commodities

### Why?

- We noticed that the commodities were getting fetched for every measure unit evaluator we were
  instantiating (quite a few of these).
- We don't want to make additional network requests if we can help it and we don't want to create a burden on the
  running application by adding more and more commodities on a colocated cache.
- We can avoid handling TTL functionality this way.

### DIFF
--- a/app/controllers/wizard/steps/base_controller.rb
+++ b/app/controllers/wizard/steps/base_controller.rb
@@ -9,6 +9,7 @@ module Wizard
       default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
       after_action :track_session
       before_action :ensure_session_integrity
+      before_action :initialize_commodity_context_service
 
       helper_method :commodity_code,
                     :commodity_source,
@@ -46,14 +47,6 @@ module Wizard
         params[:referred_service] || user_session.commodity_source
       end
 
-      def default_filter
-        { 'filter[geographical_area_id]' => country_of_origin_code }
-      end
-
-      def default_query
-        { 'as_of' => (user_session.import_date || Time.zone.today).iso8601 }
-      end
-
       def track_session
         ::NewRelic::Agent.add_custom_attributes({
           session: user_session.session.to_h.except('_csrf_token'),
@@ -74,10 +67,8 @@ module Wizard
         return redirect_to trade_tariff_url if commodity_code.blank?
       end
 
-      def country_of_origin_code
-        return user_session.other_country_of_origin if user_session.country_of_origin == 'OTHER'
-
-        user_session.country_of_origin
+      def initialize_commodity_context_service
+        Thread.current[:commodity_context_service] = CommodityContextService.new
       end
     end
   end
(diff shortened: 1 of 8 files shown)


app/controllers/wizard/steps/base_controller.rb
```diff
@@ -46,14 +47,6 @@ def commodity_source
         params[:referred_service] || user_session.commodity_source
       end
 
-      def default_filter
```
willfish, line 49: These methods were unused I noticed

app/services/expression_evaluators/measure_unit.rb
```diff
@@ -36,7 +36,7 @@ def measure_unit_answers
     end
 
     def measure_applicable_units
-      commodity.applicable_measure_units.select do |_unit, values|
+      filtered_commodity.applicable_measure_units.select do |_unit, values|
```
willfish, line 39: This reduces the size of the payload returned by the backend significantly for this request because we are only returning the commodity with the measures that apply to our origin geographical area.

willfish: Behaviourally it's exactly the same

app/controllers/wizard/steps/base_controller.rb
```diff
@@ -74,10 +67,8 @@ def ensure_session_integrity
         return redirect_to trade_tariff_url if commodity_code.blank?
       end
 
-      def country_of_origin_code
-        return user_session.other_country_of_origin if user_session.country_of_origin == 'OTHER'
-
-        user_session.country_of_origin
+      def initialize_commodity_context_service
+        Thread.current[:commodity_context_service] = CommodityContextService.new
       end
```
tavy88, line 72: So this was duplicated ?

app/helpers/commodity_helper.rb
```diff
@@ -1,20 +1,18 @@
...
 
   def commodity
-    @commodity ||= Api::Commodity.build(
-      user_session.commodity_source,
-      user_session.commodity_code,
-      default_query,
-    )
+    commodity_source = user_session.commodity_source
+    commodity_code = user_session.commodity_code
+    query = default_query
```
tavy88, line 13: Do we need this assignment?

spec/services/commodity_context_service_spec.rb
```diff
@@ -0,0 +1,44 @@
+describe CommodityContextService do
+  subject(:service) { described_class.new }
+
+  let(:commodity_source) { 'uk' }
+  let(:commodity_code) { '0702000007' }
+  let(:query) { { 'as_of' => '2021-01-01', 'filter[geographical_area_id]' => 'AR' } }
+
+  it 'generates a sha of the query' do
+    allow(Digest::SHA2).to receive(:hexdigest).with(query.to_json)
```
tavy88, line 9: Do we need to stub this?

tavy88:
Can we now remove the CommodityHelper from the steps?

tavy88 (approved):

PR #193, opened 2021-06-23
Hott 743

mafonso:
### Jira link

HOTT-743

### What?

I have added/removed/altered:

- [ ] Adds a Dockerfile to build a Docker image of Duty-calculator
- [ ] Adds a local use docker-compose to run the local version of code against the DEV api
- [ ] Adds initial CI building steps.

### Why?

I am doing this because:
- We want to have a portable docker image of the application that we can run both locally and on the live infrastructure

### DIFF
--- a/app/services/expression_evaluators/measure_unit.rb
+++ b/app/services/expression_evaluators/measure_unit.rb
@@ -36,7 +36,7 @@ module ExpressionEvaluators
     end
 
     def measure_applicable_units
-      commodity.applicable_measure_units.select do |_unit, values|
+      filtered_commodity.applicable_measure_units.select do |_unit, values|
         values['measure_sids'].include?(measure.id)
       end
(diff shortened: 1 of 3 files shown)


app/services/expression_evaluators/measure_unit.rb
```diff
@@ -36,7 +36,7 @@ def measure_unit_answers
     end
 
     def measure_applicable_units
-      commodity.applicable_measure_units.select do |_unit, values|
+      filtered_commodity.applicable_measure_units.select do |_unit, values|
```
mafonso, line 39: I don't know where this came from. A rebase with master was done, but that should not add a change against it.

Should this be removed from this PR?

tavy88: Yes please

PR #196, opened 2021-06-28
Extract measure amounts from both XI & UK services

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [ ] Extract measure amounts from both XI & UK services

### Why?

I am doing this because:

- This is needed only for RoW to NI journey, where
we need to do the deltas between XI and UK duties.

### DIFF
--- a/app/services/expression_evaluators/measure_unit.rb
+++ b/app/services/expression_evaluators/measure_unit.rb
@@ -68,7 +68,7 @@ module ExpressionEvaluators
     end
 
     def xi_measure_amounts
-      @xi_measure_amounts ||= filtered_commodity(source: 'uk').applicable_measure_units
+      @xi_measure_amounts ||= filtered_commodity(source: 'xi').applicable_measure_units
     end
   end
 end
--- a/app/services/expression_evaluators/measure_unit.rb
+++ b/app/services/expression_evaluators/measure_unit.rb
@@ -36,7 +36,13 @@ module ExpressionEvaluators
     end
 
     def measure_applicable_units
-      filtered_commodity.applicable_measure_units.select do |_unit, values|
+      unless user_session.deltas_applicable?
+        return filtered_commodity.applicable_measure_units.select do |_unit, values|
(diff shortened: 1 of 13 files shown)


app/services/expression_evaluators/measure_unit.rb
```diff
@@ -49,11 +55,20 @@ def duty_amount_in_eur?
...
+          uk_amounts
+        end
+    end
+
+    def uk_measure_amounts
+      @uk_measure_amounts ||= filtered_commodity(source: 'uk').applicable_measure_units
+    end
+
+    def xi_measure_amounts
+      @xi_measure_amounts ||= filtered_commodity(source: 'uk').applicable_measure_units
```
tavy88, line 71: This needs to be xi here
```suggestion
      @xi_measure_amounts ||= filtered_commodity(source: 'xi').applicable_measure_units
```

spec/decorators/confirmation_decorator_spec.rb
```diff
@@ -3,26 +3,31 @@
...
 
-  let(:commodity) { double(Uktt::Commodity) }
+  let(:commodity) do
+    Api::Commodity.build(
+      commodity_source,
+      commodity_code,
+    )
+  end
   let(:commodity_code) { '0702000007' }
   let(:commodity_source) { 'uk' }
```
tavy88, line 13: We don't need these anymore.

spec/decorators/confirmation_decorator_spec.rb
```diff
@@ -3,26 +3,31 @@
...
+      :with_additional_codes,
+      :with_vat,
+    )
   end
 
-  before do
-    allow(commodity).to receive(:code).and_return(commodity_code)
-    allow(Api::Commodity).to receive(:build).with(commodity_source.to_sym, commodity_code).and_return(commodity)
-  end
+  let(:referred_service) { 'uk' }
```
tavy88, line 30: We don't need this either.

willfish (approved):

PR #197, opened 2021-06-30
Enable production docker releases

mafonso:
### Jira link

HOTT-743

### What?

I have added/removed/altered:

- [ ] Enables docker releases on staging and production environments

### Why?

I am doing this because:

- We staged this change in the dev environment for over a week, during which multiple releases were done without any issues.
- This PR removes the dual logic docker/buildpack and enables docker releases across all environments

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,7 +112,7 @@ jobs:
 
   checking:
     docker:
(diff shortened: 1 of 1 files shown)


willfish:
@dependabot rebase

willfish (approved):

PR #200, opened 2021-07-01
Use merged measure units

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-776

### What?

I have added/removed/altered:

- [x] Make sure we fetch the merged measure units when deciding which ones apply in the delta case
- [x] Move methods that are shared/apply to the commodity to the commodity helper
- [x] Adds specs for the moved methods

### Why?

I am doing this because:

- This is currently broken when the source commodity doesn't have applicable measure units but the other commodity does

### DIFF
--- a/spec/helpers/commodity_helper_spec.rb
+++ b/spec/helpers/commodity_helper_spec.rb
@@ -204,7 +204,7 @@ RSpec.describe CommodityHelper do
       }
     end
 
-    context 'when on deltas are applicable' do
+    context 'when deltas are applicable' do
       let(:user_session) { build(:user_session, :with_commodity_information, :deltas_applicable) }
 
       it 'fetches the uk commodity' do
@@ -224,7 +224,7 @@ RSpec.describe CommodityHelper do
       end
     end
 
-    context 'when on deltas are not applicable' do
+    context 'when deltas are not applicable' do
       let(:user_session) { build(:user_session, :with_commodity_information) }
 
       it 'fetches the uk commodity' do
--- a/spec/helpers/commodity_helper_spec.rb
+++ b/spec/helpers/commodity_helper_spec.rb
@@ -188,4 +188,68 @@ RSpec.describe CommodityHelper do
       expect(helper.applicable_additional_codes).to eq(expected_options)
     end
   end
+
+  describe '#applicable_measure_units' do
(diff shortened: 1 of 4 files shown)


spec/helpers/commodity_helper_spec.rb
```diff
@@ -188,4 +188,68 @@
...
+          'measurement_unit_code' => 'DTN',
+          'measurement_unit_qualifier_code' => '',
+          'unit' => 'x 100 kg',
+          'unit_hint' => 'Enter the value in decitonnes (100kg)',
+          'unit_question' => 'What is the weight of the goods you will be importing?',
+        },
+      }
+    end
+
+    context 'when on deltas are applicable' do
```
willfish, line 207:
```suggestion
    context 'when deltas are applicable' do
```

spec/helpers/commodity_helper_spec.rb
```diff
@@ -188,4 +188,68 @@
...
+
+        expect(Api::Commodity).to have_received(:build).with('xi', anything, anything)
+      end
+
+      it 'fetches the measure units' do
+        expect(helper.applicable_measure_units).to eq(expected_measure_units)
+      end
+    end
+
+    context 'when on deltas are not applicable' do
```
willfish, line 227:
```suggestion
    context 'when deltas are not applicable' do
```

spec/helpers/commodity_helper_spec.rb
```diff
@@ -188,4 +188,68 @@
...
+      }
+    end
+
+    context 'when deltas are applicable' do
+      let(:user_session) { build(:user_session, :with_commodity_information, :deltas_applicable) }
+
+      it 'fetches the uk commodity' do
+        helper.applicable_measure_units
+
+        expect(Api::Commodity).to have_received(:build).with('uk', anything, anything)
```
tavy88, line 213: Is `anything` this a reserved keyword?

willfish: It instantiates a matcher of the type RSpec::Mocks::ArgumentMatchers::AnyArgMatcher. Basically it's a method call that returns a AnyArgMatcher which makes any argument of any type pass comparison

tavy88 (approved):

PR #201, opened 2021-07-01
Collect additional code answers from both uk & xi

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] When there deltas apply on the RoW to NI journey, we need
to fetch the additional code options from both UK and XI
services as we will need to use the selected options
in the duty calculation later on.

### Why?

I am doing this because:

- Needed for the RoW to NI jounrey

### DIFF
--- a/app/models/wizard/steps/additional_code.rb
+++ b/app/models/wizard/steps/additional_code.rb
@@ -6,29 +6,32 @@ module Wizard
       STEPS_TO_REMOVE_FROM_SESSION = %w[].freeze
 
       attribute :measure_type_id, :string
-      attribute :additional_code, :string
+      attribute :additional_code_uk, :string
+      attribute :additional_code_xi, :string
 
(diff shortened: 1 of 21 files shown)


app/models/wizard/steps/additional_code.rb
```diff
@@ -60,27 +65,33 @@ def build_option(code, overlay)
...
-        available_measure_types[previous_measure_type_index]
+        applicable_measure_type_ids[previous_measure_type_index]
       end
 
-      def available_measure_types
-        @available_measure_types ||= applicable_additional_codes.keys
-      end
+      # def available_measure_types
+      #   @available_measure_types ||= (applicable_additional_codes['uk'].keys + applicable_additional_codes['xi'].keys).uniq
+      # end
```
tavy88, line 81: Must go.

willfish (approved):

PR #210, opened 2021-07-07
HOTT-783: DC - Measure Amount with more than 5 digits after decimal showing in scientific notation 1.0e-05

cacquani:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-783
https://transformuk.atlassian.net/browse/HOTT-787

### What?

I have added/removed/altered:

- [ ] Format the amount so that it will be displayed in human readable notation up to 10 decimals
- [ ] Make sure numbers are displayed with at least two decimals even when they start as integers
- [ ] Add specs

### Why?

I am doing this because this is a site for traders and scientific notation is hard to read unless you're in the scientific field and very much used to it.

### Have you? (optional)


spec/services/expression_evaluators/measure_unit_spec.rb
```diff
@@ -32,43 +32,87 @@
...
+
+      it 'returns a properly calculated evaluation' do
+        expect(evaluator.call).to eq(expected_evaluation)
+      end
+    end
   end
 
-  context 'when on a route with applicable deltas' do
-    let(:user_session) { build(:user_session, :deltas_applicable, session_attributes) }
+  context 'with a quantity with more four decimals' do
```
willfish, line 77: I think if you delegate to the number_to_precision helper with whatever locale we've setup you don't really need a dedicated unit test for this?

app/services/duty_options/base.rb
```diff
@@ -63,7 +63,14 @@ def measure_unit?
     end
 
     def total_quantity
-      number_with_precision(duty_evaluation[:total_quantity], precision: 10, significant: false, strip_insignificant_zeros: true)
+      if /[\.e]/ !~ duty_evaluation[:total_quantity].to_s || /\.\d$/ =~ duty_evaluation[:total_quantity].to_s
```
willfish, line 66: How do you feel about moving this to a reusable service?

```ruby

class MeasureComponentQuantityFormatService
  include ActionView::Helpers::NumberHelper

  def initialize(quantity)
    @quantity = quantity
  end

  def call
    number_with_precision(
      @quantity,
      precision: precision,
      significant: false,
      strip_insignificant_zeros: strip_zeroes,
    )
  end

  private

  def precision
    return 10 if scientific_notation?

    2
  end

  def strip_insignificant_zeros
    return true if scientific_notation?

    false
  end

  def scientific_notation?
    @scientific_notation ||= /[.e]/ =~ quantity.to_s || /\.\d$/ !~ quantity.to_s
  end
end
```

We can then write unit tests more easily and just consume it in the duty options and the evaluator?

cacquani: Sounds like a good idea, but I would keep the name more generic - it may end up being useful for other quantities other than the measure amount.
How does `NumberWithHighPrecisionFormatter` sound?

willfish: Sounds perfect :)

app/services/number_with_high_precision_formatter.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def strip_insignificant_zeros
+    return false if low_precision? && !scientific_notation?
+
+    true
+  end
+
+  def scientific_notation?
+    /e/.match? @quantity.to_s
```
willfish, line 34: Nice :)

willfish (approved):

PR #224, opened 2021-07-12
Remove rails and load the gems we actually use

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes canonical rails - this was inherited from another project and we're not actually using this.
- [x] Use specific gems with functionality we actually use rather than just loading all of rails

### Why?

I am doing this because:

- Lighten up the project somewhat and reduce update/security burdens
- We specifically don't want the search engines to know about our application.


cacquani (approved):
The changes look good to me

PR #246, opened 2021-07-20
Plumbs together excise step

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-832

### What?

The excise step is quite similar to the additional code step. It uses
slightly different namespacing for the session and filters on different
additional codes. We may want to consider a refactor separately to DRY up the two steps/reuse behaviour as much as possible.

Excise needs to _always_ pull from the uk data source (like VAT)

This adds:

- [x] The excise step model for validating and persisting the session
  attributes
- [x] The excise controller
- [x] The meaning of an excise measure type in the measure type api model
- [x] The excise show view
- [x] The excise next/previous step paths
- [x] Locale strings for error message validations
- [x] Specs covering the new step and related classes
- [x] Adds confirmation page plumbing

### Why?

I am doing this because:

- I've gone for the full list of supported measure types now since I know I'm going to need them later down the line.
- I've avoided sharing behaviour because the concepts are similar but have separate copy, session keys, etc. If we decide add more measure-filtering additional code steps in future then I definitely think the complexity/DRY battle will be won for sharing behaviour between the steps


willfish:
image

willfish:
image

willfish:
image

app/models/steps/excise.rb
```diff
@@ -0,0 +1,74 @@
...
+  class Excise < Steps::Base
+    STEPS_TO_REMOVE_FROM_SESSION = %w[].freeze
+
+    include CommodityHelper
+
+    attribute :measure_type_id, :string
+    attribute :additional_code, :string
+
+    validates :measure_type_id, presence: true
+    validates :additional_code, presence: true
```
jebw, line 11: Does there need to be additional validation around the codes supplied by the user?

jebw (commented):
Looks good but couple of questions around the method on UserSession and wondering if theres any way to validate the additional_codes supplied by the user to the Excise model?

app/models/user_session.rb
```diff
@@ -205,6 +218,10 @@ def additional_codes
     (additional_code_uk.values + additional_code_xi.values).compact.join(', ')
   end
 
+  def excise_additional_codes
```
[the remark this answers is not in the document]
willfish: Sounds good :+1:

app/models/steps/excise.rb
willfish: Talked this through. :+1:

jebw (approved):

PR #250, opened 2021-07-21
Move to using monetary exchange rate

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-855

### What?

I have added/removed/altered:

- [x] Move to using monetary exchange rates from the XI service

### Why?

I am doing this because:

- This is a strict HMRC requirement


Gemfile
```diff
@@ -16,7 +16,7 @@ gem 'newrelic_rpm'
 gem 'puma'
 gem 'railties', '~> 6'
 gem 'sentry-raven'
-gem 'uktt', git: 'https://github.com/trade-tariff/uktt.git'
+gem 'uktt', git: 'https://github.com/trade-tariff/uktt.git', ref: 'monetary_exchange_rates'
```
willfish, line 19: When I've merged the UKTT gem branch this can be updated to not include the ref

jebw (approved):

PR #251, opened 2021-07-21
Refactor user_session

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Stop passing the user session everywhere

### Why?

I am doing this because:

- We currently pass the user session everywhere in the context of the
current request. This is painful and creates a lot of noise.


app/models/user_session.rb
```diff
@@ -213,6 +213,14 @@ def no_duty_to_pay?
     no_duty_route? || possible_duty_route? && no_duty_applies?
   end
 
+  def self.build(session)
+    Thread.current[:user_session] ||= UserSession.new(session)
```
willfish, line 217:
```suggestion
    Thread.current[:user_session] || = new(session)
```

jebw (approved):

PR #259, opened 2021-07-26
Show chosen document codes on confirmation page

tavy88:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-676

### What?

I have added/removed/altered:

- [x] Show chosen document codes on the confirmation page

### Why?

I am doing this because:

- users need to be able to change their selection
- users need to see a summary of their selection

Screen Shot 2021-07-26 at 11 06 25

### DIFF
--- a/app/decorators/confirmation_decorator.rb
+++ b/app/decorators/confirmation_decorator.rb
@@ -58,7 +58,7 @@ class ConfirmationDecorator < SimpleDelegator
     return format_measure_amount(value) if key == 'measure_amount'
     return country_name_for(value, key) if %w[import_destination country_of_origin].include?(key)
     return additional_codes_for(value) if key == 'additional_code'
-    return document_code_for(value) if key == 'document_code'
+    return selected_document_codes_from(value) if key == 'document_code'
     return excise_for(value) if key == 'excise'
     return vat_label(value) if key == 'vat'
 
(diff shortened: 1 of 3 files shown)


app/decorators/confirmation_decorator.rb
```diff
@@ -110,6 +113,20 @@ def excise_additional_codes
     user_session.excise_additional_code.values.join(', ')
   end
 
+  def additional_codes
+    (user_session.additional_code_uk.values + user_session.additional_code_xi.values).compact.join(', ')
+  end
+
+  def document_code_for(value)
+    return nil unless value.values.reject(&:empty?).any?
```
jebw, line 121: Really minor code style but this line isn't the most straight forward to read? An `if` would remove the double negative and renaming the method arg could make it clearer?

```ruby
def document_code_for(????)
  return nil if ????.values.reject(&:empty?).empty?

  document_codes
end

jebw (approved):
LGTM but suggestion for minor tweak for code readability?

app/decorators/confirmation_decorator.rb
tavy88: Valid points! Will update.

jebw (approved):
LGTM :+1:

PR #262, opened 2021-07-28
Return n/a if no document is selected

tavy88:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] When on the confirmation page, if the user has not
selected a document code, we just display n/a.

### Why?

I am doing this because:

- UX request

### DIFF
--- a/spec/decorators/confirmation_decorator_spec.rb
+++ b/spec/decorators/confirmation_decorator_spec.rb
@@ -70,6 +70,46 @@ RSpec.describe ConfirmationDecorator, :user_session do
       ]
     end
 
+    context 'when no document code has been selected' do
+      let(:user_session) do
+        build(
+          :user_session,
+          :with_commodity_information,
+          :with_import_date,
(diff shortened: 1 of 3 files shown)


spec/decorators/confirmation_decorator_spec.rb
```diff
@@ -70,6 +70,46 @@
...
+          :with_import_date,
+          :with_import_destination,
+          :with_country_of_origin,
+          :with_trader_scheme,
+          :with_certificate_of_origin,
+          :with_customs_value,
+          :with_measure_amount,
+          :with_additional_codes,
+          :with_excise_additional_codes,
+          :with_vat,
```
willfish, line 87: We could jettison all of these other traits and just focus on the document codes?

tavy88: Are you very precious about it?

willfish (approved):

PR #263, opened 2021-07-29
Handle sometimes present suspension_legal_act

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Adds handling for sometimes present legal act
- [x] Adds handling for VAT measure that does not have an additional code

### Why?

I am doing this because:

- https://sentry.io/organizations/engine-le/issues/2540682038/?referrer=slack

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,7 @@ gem 'newrelic_rpm'
 gem 'puma'
 gem 'railties', '~> 6'
 gem 'sentry-raven'
-gem 'uktt', git: 'https://github.com/trade-tariff/uktt.git'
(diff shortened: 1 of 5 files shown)


willfish:
@tavy88 Approved verbally

jebw (approved):
Code looks great, pretty clear what its doing - there's no changes to specs though?

I can't find a spec for `models/api/base.rb`? I think adding some specs for the basic behaviour around attributes and relationships (with a mock model instance?) 

```ruby
let(:TestApiModel) do
  Class.new(Api::Base) do
      # model implementation
  end
end
```

Specs for the 'real' models instances which just check for which attributes were on them would help catch things like typo's in attr names, etc ?

PR #267, opened 2021-08-02
Apply duty calculations for document codes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-677

### What?

The purpose of this PR is to enable handling document codes that come from the tariff api and that traders need to have in order to take advantage of a measure or avoid being punished by some punitive additional duty.

We need to take answers about the document code that applies for a given measure type and select the condition that applies for that document code. What we do with the document code is determined by the action code for the applicable document measure condition. These conditions contain components that we can then pass to the duty calculator process to apply the mentioned duty or (in other cases) the document code/lack thereof requires that we need to exclude the measure completely (for instance you need a certificate in order to take advantage of a preferential tariff). 

You can see some good examples, here: https://dev.trade-tariff.service.gov.uk/commodities/1516209821#import

**Apply the mentioned duty (fall back to the duty expressed on the measure's measure components) or skip this measure altogether (if answered in the negative)**

Screenshot from 2021-08-02 13-47-49

**Apply one of the mentioned measure condition's duties**

Screenshot from 2021-08-02 13-48-11

I have added/removed/altered:

- [x] Document codes are exclusive per measure type (move from check boxes to radio buttons)
- [x] Move the understanding/business logic of the applicable measures for a given commodity to the commodity model
- [x] Expose the user session which contains the document code answers to the api models
- [x] Migrates getters of session values directly into the session wrapper class
- [x] Handle deltas applicable route when determining the measure type ids that apply to our route (fix the same issue in passing with the additional codes)
- [x] Handle 'Apply the mentioned duty' condition actions and pass the related components to the duty calculation
- [x] Handle ''Measure not applicable' measure actions

### Why?

I am doing this because:

- As set out above this is needed to apply the result of a trader having/not having a certain document


app/services/expression_evaluators/measure_unit.rb
```diff
@@ -71,5 +78,15 @@ def uk_measure_amounts
     def xi_measure_amounts
       @xi_measure_amounts ||= filtered_commodity(source: 'xi').applicable_measure_units
     end
+
+    def measure_condition
+      return nil if component.is_a? Api::MeasureComponent
+
+      measure.measure_conditions.detect do |measure_condition|
+        measure_condition.measure_condition_components.any? do |measure_condition_component|
+          measure_condition_component.as_json == component.as_json
```
jebw, line 87: Casting to a json string seems an odd way of determining if its the measure condition component your looking for? Could this be a redefining of `==` on MeasureConditionComponent? and a more defined comparison on that?

app/views/steps/document_codes/show.html.erb
```diff
@@ -10,11 +10,31 @@
...
 
   <% if @step.applicable_document_codes.dig('uk', @step.measure_type_id).present? %>
-    <%= f.govuk_collection_check_boxes :document_code_uk, @step.options_for_checkboxes_for(source: 'uk'), :id, :name, legend: { text: '' } %>
+    <%= f.govuk_collection_radio_buttons(
+      :document_code_uk,
+      @step.options_for_radio_buttons(source: 'uk'),
+      :id,
+      :name,
+      legend: { text: '' },
+      hint: {
```
jebw, line 19: can hint and legend follow the same linting pattern?

app/views/steps/document_codes/show.html.erb
```diff
@@ -10,11 +10,31 @@
...
-    <%= f.govuk_collection_check_boxes :document_code_xi, @step.xi_options_for_checkboxes_without_uk, :id, :name, legend: { text: '' } %>
+  <% if @step.xi_options_for_radio_buttons_without_uk.present? %>
+    <%= f.govuk_collection_radio_buttons(
+      :document_code_xi,
+      @step.options_for_radio_buttons(source: 'xi'),
+      :id,
+      :name,
+      legend: { text: '' },
+      hint: {
+        text: ''
```
jebw, line 34: can hint and legend follow the same linting pattern?

spec/controllers/steps/document_codes_controller_spec.rb
```diff
@@ -26,16 +26,17 @@
     end
 
     context 'when the step answers are valid' do
-      let(:document_code_uk) { ['C644', 'Y929', ''] }
-      let(:document_code_xi) { ['N851', ''] }
+      let(:document_code_uk) { 'C990' }
```
jebw, line 29: Differeny codes would help check their operating independently?

spec/models/api/commodity_spec.rb
```diff
@@ -1,8 +1,9 @@
-RSpec.describe Api::Commodity, type: :model do
+RSpec.describe Api::Commodity, :user_session, type: :model  do
```
jebw, line 1: super minor but the extra whitespace after `:model` is not needed

jebw (approved):
Looks great - bunch of minor opinions on naming/code structure but take or leave as relevant

Only point point that stands out is `all_measures` which looks like it isn't all measures since its a conditional of one source or another? That may be me misunderstanding it though

One request would be if there are any other changes to the branch before merge - can you avoid any rebases until right before the actual merge - this is so big I've got no chance of determining the changes made otherwise.

Thanks

app/services/expression_evaluators/measure_unit.rb
willfish: As json is fine but I could move the location of the comparison with an override of the == method in the measure_condition_component model

app/models/api/measure.rb
```diff
@@ -80,8 +79,31 @@ def vat_type
...
+    end
+
     private
 
+    def document_components
+      applicable_document_condition&.measure_condition_components
+    end
+
+    def applicable_document_condition
+      @applicable_document_condition ||= begin
```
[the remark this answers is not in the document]
willfish: I mean it is all in memory and it's not a lot of work to iterate through measure conditions and pick out the one that matches a value stored in the user session hash but I don't see any harm in the computer not doing it all again.

app/models/api/measure.rb
```diff
@@ -63,8 +63,7 @@ def component
     end
 
     def all_components
-      # TODO: This needs to include measure condition components
-      @all_components ||= measure_components
+      @all_components ||= document_components.presence || measure_components
```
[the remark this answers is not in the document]
willfish: I think it makes sense. Look in all the place places for all possible components of either type (measure condition component or measure component). A more concrete name might be `all_measure_components_or_measure_condition_components` or simply `components`

Whatever reflects the fact that we want potentially multiple components that might come from different places.

app/models/api/commodity.rb
```diff
@@ -43,5 +43,28 @@ def formatted_commodity_code(additional_code = nil)
...
+    private
+
+    def no_additional_code_measures
+      import_measures_no_vat.reject(&:additional_code)
+    end
+
+    def additional_code_measures
+      import_measures_no_vat.select do |measure|
+        additional_code = measure.additional_code
+        code_answer = user_session.additional_code_for(measure.measure_type.id, source)
```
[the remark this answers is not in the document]
willfish: Have opted for:

```ruby
        measure_additional_code = measure.additional_code
        user_additional_code_answer = user_session.additional_code_for(measure.measure_type.id, source)

        measure_additional_code.present? && user_additional_code_answer == measure_additional_code.code
```

PR #277, opened 2021-08-06
Duty calculations for excise measures

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-834

### What?

I have added/removed/altered:

- [x] Integrate excise into duty calculation page
- [x] Adds corresponding unit tests for the new option

image

### Why?

I am doing this because:

- This is a business requirement for the duty calculator

### Acceptance Criteria

- Handles excise classes in https://transformuk.atlassian.net/browse/HOTT-827
- Handles excise classes in https://transformuk.atlassian.net/browse/HOTT-828
- Handles excise classes in https://transformuk.atlassian.net/browse/HOTT-829

### DIFF
--- a/app/models/api/additional_code.rb
+++ b/app/models/api/additional_code.rb
@@ -1,9 +1,17 @@
 module Api
   class AdditionalCode < Api::Base
+    EXCISE_ADDITIONAL_CODE_TYPE_ID = 'X'.freeze
+
     attributes :id,
                :code,
                :description,
                :formatted_description
+
+    def formatted_code
+      return additional_code_id if excise?
+
+      code
+    end
 
     def additional_code_type
       code && code[0]
@@ -11,6 +19,10 @@ module Api
 
     def additional_code_id
       code && code[1..]
+    end
+
+    def excise?
+      additional_code_type == EXCISE_ADDITIONAL_CODE_TYPE_ID
     end
   end
 end
--- a/app/models/api/additional_code.rb
+++ b/app/models/api/additional_code.rb
@@ -1,11 +1,13 @@
 module Api
   class AdditionalCode < Api::Base
-    EXCISE_ADDITIONAL_CODE_TYPE_ID = 'X'.freeze
-
     attributes :id,
                :code,
                :description,
                :formatted_description
+
+    enum :additional_code_type, {
+      excise: %w[X],
+    }
 
     def formatted_code
       return additional_code_id if excise?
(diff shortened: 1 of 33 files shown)


app/models/api/additional_code.rb
```diff
@@ -1,16 +1,28 @@
 module Api
   class AdditionalCode < Api::Base
+    EXCISE_ADDITIONAL_CODE_TYPE_ID = 'X'.freeze
```
willfish, line 3: Probably this would be better served by using the enum api :thinking:

spec/services/duty_options/quota/non_preferential_end_use_spec.rb
```diff
@@ -14,7 +14,7 @@
         value: 96,
         order_number: '058048',
         measure_sid: measure.id,
-        source: nil,
+        source: 'uk',
```
willfish, line 17: If you're curious, this changed because I was properly serializing the measures in the factory (`measure#source`) gets called so we can know where the calculation comes from for a given duty option

app/services/duty_options/additional_duty/excise.rb
```diff
@@ -0,0 +1,26 @@
...
+      PRIORIY = 6
+
+      include CommodityHelper
+
+      def duty_calculation_row
+        presented_row = []
+        presented_row << I18n.t(
+          'duty_calculations.options.excise_duty_html',
+          additional_code_description: excise_description,
+        ).html_safe
```
jebw, line 13: very minor but I don't think `.html_safe`  is needed - a `_html` key will get marked html_safe automatically

app/services/duty_options/additional_duty/excise.rb
```diff
@@ -0,0 +1,26 @@
...
+
+        presented_row.concat(duty_evaluation.slice(:calculation, :formatted_value).values)
```
jebw, line 15: Again really minor but I think `#values_at` does what you want here ?

jebw (approved):
LGTM - couple of really minor comments

app/services/duty_options/additional_duty/excise.rb
willfish: Nice

app/services/duty_options/additional_duty/excise.rb
willfish: Feeexed

PR #278, opened 2021-08-06
Extend deltas applicable logic to exclude trade defence measures

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-884

### What?

I have added/removed/altered:

- [x] Force EU measures when there are trade defence measures in place on the deltas applicable route
- [x] Extend user session factory to define the deltas applicable logic to fix the tests

### Why?

I am doing this because:

- EU duties always apply whenever there are trade defence measures on the EU measures for a given commodity

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -83,6 +83,8 @@ class UserSession
     session['trade_defence']
   end
 
+  alias_method :trade_defence?, :trade_defence
+
   def trade_defence=(value)
     session['trade_defence'] = value
   end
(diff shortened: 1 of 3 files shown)


jebw (changes requested):
This should have unit tests around `delta's_applicable?`

willfish:
I've changed the formatting/descriptions of the deltas applicable specs to make it obvious they're there @jebw

jebw (approved):

PR #282, opened 2021-08-10
Handle duplicated constants causing warnings

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Make it so that both MeasureComponent and MeasureConditionComponent have exactly conforming implementations
- [x] Test the base implementation with shared behaviour and check
inheritance, only.

### Why?

I am doing this because:

- To remove the warning about duplicate constants being loaded for the mathematical operators
- To reduce implementation creep between the two ducks.
- To reduce test duplication


jebw (approved):
LGTM, cleans up a lot - only thought would be using a mixin/concern instead of parent class (composition vs inheritance)? You'd need to use an anonymous class for the Api::Base spec then though so definite tradeoffs

willfish:
Yeah, definitely doing the mixin a lot and finding it painful for other reasons :confused:

PR #283, opened 2021-08-10
Format the base expression

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Move merging of the deltas applicable measure units into a service
- [x] Test new service in isolation
- [x] Use new service in commodity helper
- [x] Test using new service in commodity helper and remove old deltas applicable test that's not adding lots of value
- [x] Updates all duty expressions with units with spans and abbr tags to show them
- [x] Move to using the measure and duty expression factories wherever possible
- [x] Updates acceptance level tests of the duty calculator to indicate the change of formatting

Before

Screenshot from 2021-08-10 15-35-02

After

Screenshot from 2021-08-10 15-23-35

### Why?

I am doing this because:

- I'll need this merging service later but it's definitely much easier to test this way and I was touching the code to change the formatting.
- The formats could be improved for the end user

### DIFF
--- a/app/services/expression_evaluators/compound.rb
+++ b/app/services/expression_evaluators/compound.rb
@@ -2,13 +2,17 @@ module ExpressionEvaluators
   class Compound < ExpressionEvaluators::Base
     def call
       {
-        calculation: measure.duty_expression.base,
+        calculation: calculation_duty_expression,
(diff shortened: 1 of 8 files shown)


jebw (approved):
LGTM

PR #285, opened 2021-08-11
Bumps uktt

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Bumps uktt gem

### Why?

I am doing this because:

- We're moving to using persistent faraday for better performance
- We're also validating/public/private routing of the uktt gem after the change


jebw (approved):
:ok_hand:

willfish:
Not a terrible idea :). I'd be tempted to start by just reporting the status/response, etc, to sentry (I think that might already happen, actually)

PR #287, opened 2021-08-12
Handle missing excise measures on XI routes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-888

### What?

I have added/removed/altered:

- [x] Enable different xi/uk fixtures. Remove unused fixtures.
- [x] Extend ApplicableMeasureUnitMerger to enable handling of delta, xi and uk routes
- [x] Adds AdditionalDutyApplicableMeasuresMerger to enable merging of additional duty rows when on xi routes in the duty calculator. Previously applicable measures for additiona options and normal options would only come from one service. Now we have excise measures coming from the uk service, only and being merged into the measures on the xi route.
- [x] Adds missing test coverage
- [x] Adds missing factories

### Why?

I am doing this because:

- This is needed for the duty calculator to ask for the proper measure units about excise when on the xi routes
- This is needed to do duty calculations on additional duty options on the xi route


app/services/additional_duty_applicable_measures_merger.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def xi_applicable_measures
+    filtered_commodity.applicable_measures.select(&:applicable?).concat(uk_filtered_commodity.excise_measures)
+  end
+
+  def uk_applicable_measures
+    filtered_commodity.applicable_measures.select(&:applicable?)
+  end
+
+  def user_session
```
willfish, line 18: Planning on rejigging this so it isn't needed as much - there seems to be a significant crossover with the commodity helper so I can move this method there and kill it in like 5 other places

app/models/api/commodity.rb
```diff
@@ -48,8 +48,28 @@ def applicable_measures
       @applicable_measures ||= no_additional_code_measures + additional_code_measures
     end
 
+    def applicable_excise_measure_units
+      applicable_measure_units.each_with_object({}) do |(unit, value), acc|
```
jebw, line 52: Presumably `acc` is an abbreviation for accumulator - full name would be easier to read though?

app/models/api/commodity.rb
```diff
@@ -48,8 +48,28 @@ def applicable_measures
...
+        measure_sids = value['measure_sids'].select { |measure_sid| measure_sid.in?(excise_measures_sids) }
+
+        next if measure_sids.blank?
+
+        value.deep_dup['measure_sids'] = measure_sids
```
jebw, line 57: I might be missing but this line looks like it does nothing to me - dup the hash, set a key, then ignore the result ?

jebw (approved):
Looks good - namespacing the JSON responses looks really beneficial.

There's a variable that could probably benefit from a clearer name? There's also a line which I think may be redundant?

:+1:

app/models/api/commodity.rb
willfish: I'll rename this separately since it's the name we have historically used for the accumulator.

PR #291, opened 2021-08-16
Removes unused feature flag for document codes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-894

### What?

I have added/removed/altered:

- [x] Enable document codes by removing the feature flag

### Why?

I am doing this because:

- We've tested and are now ready to go live


jebw (changes requested):
Needs the env vars in test and dev removing

jebw (approved):
LGTM

PR #294, opened 2021-08-19
Handle retail price measure units

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [ ] 
- [ ] Removed baz

### Why?

I am doing this because:

-
-
-

### Have you? (optional)

- Makes changes to our complex routing setup that may affect apis to proxying to backend


app/services/duty_options/base.rb
```diff
@@ -165,10 +165,6 @@ def presented_commodity_source
       measure.source == 'xi' ? 'EU' : 'UK'
     end
 
-    def uk_filtered_commodity
```
willfish, line 168: This is unused and duplicates something the same method in the commodity helper

willfish:
Screenshot from 2021-08-24 12-14-49

app/helpers/measure_unit_helper.rb
```diff
@@ -0,0 +1,12 @@
+module MeasureUnitHelper
```
jebw, line 1: Is this view helper code? Feels more like its building a data structure? - Would it make more sense as a module a `services/concerns` folder?

app/helpers/measure_unit_helper.rb
```diff
@@ -0,0 +1,12 @@
...
+  def presented_unit
+    @presented_unit ||= {
+      answer: user_session.measure_amount[component.unit.downcase.to_s],
```
jebw, line 4: This is encapsulating logic about how to get a key suitable for `measure_amount` from a `component.unit` - that logic could be moved to the data object, eg `user_session.measure_amount[component.unit_key]`

jebw (approved):
LGTM - only observation is the logic in the view helper? That could possibly be moved out to a concern?

app/helpers/measure_unit_helper.rb
willfish: Yeah, makes sense. I've moved it to a concern :)

PR #296, opened 2021-08-20
Handle measure units coming from measure components and measure condition components

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-907

### What?

I have added/removed/altered:

- [x] Adds support for pulling out measure units for a given answer and component from the components id
- [x] Extract excise measure unit finding into it's own service
- [x] Removes a now-unused fixture
- [x] Adds unit tests

### Why?

I am doing this because:

- This is needed because we want to identify the specific component a unit belongs too and not the broader measure


willfish:
Superseded by https://github.com/trade-tariff/trade-tariff-duty-calculator/pull/297

PR #297, opened 2021-08-20
Removes merging complexity in merger now that we can ignore the sids

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Handles pulling out the measure unit answers for the measure unit evaluator without needing to lookup the measure sid

### Why?

I am doing this because:

- This wasn't actually ever needed - the component tells you the unit

### DIFF
--- a/app/services/expression_evaluators/measure_unit.rb
+++ b/app/services/expression_evaluators/measure_unit.rb
@@ -7,7 +7,7 @@ module ExpressionEvaluators
         calculation: calculation_duty_expression,
         value: value,
         formatted_value: number_to_currency(value),
-        unit: measure_unit_answers.first[:unit],
+        unit: presented_unit[:unit],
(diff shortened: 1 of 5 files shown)


app/services/expression_evaluators/measure_unit.rb
```diff
@@ -40,24 +40,18 @@ def value
...
     end
 
-    def measure_applicable_units
-      units = ApplicableMeasureUnitMerger.new.call
-
-      units.select do |_unit, values|
-        values['measure_sids'].include?(measure.id)
-      end
+    def applicable_unit
+      ApplicableMeasureUnitMerger.new.call[component.unit]
```
willfish, line 54: Just pull out the unit defined by the component rather than looking up the unit by the id of either the component or the measure condition component

jebw (approved):

PR #300, opened 2021-08-24
Enable excise calculations globally

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-895

### What?

I have added/removed/altered:

- [x] Remove env var control of excise step

### Why?

I am doing this because:

- We're nearly ready to go live with excise


jebw (approved):
LGTM

PR #303, opened 2021-08-25
Removes Factor and ASV percent measure units from duty calculation consideration

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes FC1X from all routes measure units

### Why?

I am doing this because:

- We currently don't know how to handle this and the additional code filtering that would stop this from being shown isn't currently happening


jebw (approved):
LGTM

PR #307, opened 2021-08-31
Extract duty options into a duty option result model

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-809

### What?

I have added/removed/altered:

- [x] Correctly name service apis
- [x] Updates specs to reflect the change
- [x] Make sure duty option api is consistent everywhere
- [x] Updates factory to reflect new result class
- [x] Updates references to duty options
- [ ] Updates appending behaviour for suffixes if there is an appended suffix already

### Why?

I am doing this because:

- It was cleaner to refactor and accumulate into a duty option result footnote suffix than it was to generate to hack on the row calculator


jebw (approved):
LGTM - nice refactor!

PR #308, opened 2021-08-31
Accept only the first footnote suffix

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-809

### What?

I have added/removed/altered:

- [x] Accept only the first suffix in the duty option result
- [x] Adds corresponding specs for the footnote suffix behaviour

### Why?

I am doing this because:

- This is needed to avoid accumulating suffixes into the duty option result

### DIFF
--- a/app/models/duty_option_result.rb
+++ b/app/models/duty_option_result.rb
@@ -15,11 +15,11 @@ class DutyOptionResult
   attribute :geographical_area_description
 
   def footnote
-    (super + footnote_suffix).html_safe
+    super + footnote_suffix
   end
 
   def footnote_suffix
-    @footnote_suffix.presence || ''
+    @footnote_suffix.presence || ''.html_safe
   end
 
   def footnote_suffix=(suffix)
--- a/app/models/duty_option_result.rb
+++ b/app/models/duty_option_result.rb
@@ -13,4 +13,18 @@ class DutyOptionResult
   attribute :warning_text
   attribute :order_number
   attribute :geographical_area_description
+
+  def footnote
+    (super + footnote_suffix).html_safe
+  end
+
+  def footnote_suffix
+    @footnote_suffix.presence || ''
+  end
+
+  def footnote_suffix=(suffix)
+    return if @footnote_suffix
+
+    @footnote_suffix = suffix
+  end
 end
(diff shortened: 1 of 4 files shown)


app/models/duty_option_result.rb
```diff
@@ -13,4 +13,18 @@ class DutyOptionResult
   attribute :warning_text
   attribute :order_number
   attribute :geographical_area_description
+
+  def footnote
+    (super + footnote_suffix).html_safe
```
jebw, line 18: Does footnote_suffix have html content in? If not you could use safe_join which will auto escape?

```ruby
safe_join [super, footnote_suffix]
```

if it does have html in - if you mark the suffix as safe, concatenated them should leave them both mark as safe (assuming `super` returns an `html_safe` string)

app/models/duty_option_result.rb
```diff
@@ -13,4 +13,18 @@ class DutyOptionResult
...
+  end
+
+  def footnote_suffix
+    @footnote_suffix.presence || ''
```
jebw, line 22: Maybe mark it as html_safe here instead? `(@footnote_suffix.presence || '').html_safe`

app/models/row_to_ni_duty_calculator.rb
```diff
@@ -40,7 +40,7 @@ def options
 
       footnote_suffix ||= I18n.t("row_to_ni_measure_type_footnotes_suffixes.#{option.category}.#{option.source}").html_safe
 
-      option.footnote = option.footnote.concat(footnote_suffix).html_safe
+      option.footnote_suffix = footnote_suffix
```
jebw, line 43: If this is doing the assignment, this is probably the best place to be deciding if the suffix is safe?

app/models/duty_option_result.rb
jebw: `String#+` is overridden by rails - if all components are html_safe, the string remains html_safe

```ruby
('a'.html_safe + 'b'.html_safe).html_safe? # => true

('a'.html_safe + 'b').html_safe? # => false
```

jebw (approved):
Looks good but the marking of the whole footnote as always safe is potentially not needed?

PR #310, opened 2021-09-02
Adds page title overrides for steps, duties and interstitial messages

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-613

### What?

I have added/removed/altered:

- [x] Added title override for duty controller
- [x] Added title override for interstitial message controller
- [x] Added title override for step controllers
- [x] Remove unused layout templates
- [x] Remove duplicated empty step constant

### Why?

I am doing this because:

- These are needed for GDS compliance
- I've skipped writing UI tests here since the views change so frequently
- I avoided adding a dedicated helper as it was more complex than simply overriding the title helper


app/controllers/steps/base_controller.rb
```diff
@@ -65,5 +66,9 @@ def ensure_session_integrity
     def initialize_commodity_context_service
       Thread.current[:commodity_context_service] = CommodityContextService.new
     end
+
+    def title
+      t("page_titles.#{@step.class.try(:id)}", default: super)
```
alexdesi, line 71: I am wondering if we could create a method where @step is defined instead access it and call `step.class.try(:id)` here.

I just wanted to try a cleaner way, but not sure how at the moment.

app/controllers/steps/duty_controller.rb
```diff
@@ -1,5 +1,7 @@
 module Steps
   class DutyController < BaseController
+    helper_method :title
+
     def show
       @duty_options = duty_options
       @gbp_to_eur_exchange_rate = Api::MonetaryExchangeRate.for('GBP').exchange_rate.round(4) unless user_session.import_into_gb?
```
alexdesi, line 7: this means that if `user_session.import_into_gb?` is nil,
@gbp_to_eur_exchange_rate will remain nil, right?
maybe a ternary operator would be clearer

app/controllers/steps/duty_controller.rb
```diff
@@ -21,5 +23,11 @@ def uk_options
     def xi_options
       @xi_options ||= DutyCalculator.new(filtered_commodity(source: 'xi')).options
     end
+
+    def title
+      t('page_titles.duty_calculation') if @duty_options.present?
+
+      t('page_titles.no_duty')
+    end
```
alexdesi, line 31: does this method always return `t('page_titles.no_duty')`? 
I think you missed a "return"

This could go too:
```
def title
   @duty_options.present? ? t('page_titles.duty_calculation') : t('page_titles.no_duty')
end
```

willfish:
Ah good catch @alexdesi . Will fix

app/controllers/steps/base_controller.rb
willfish: As far as I can make out the only place we need to do this is here so I don't mind too much asking the step for it's class just once. I reckon if this pattern repeats a bunch we can add an instance method on the step to pull out the id of the singleton, though

app/controllers/steps/duty_controller.rb
willfish: Yeah, this is the guard clause style of basically doing:

```ruby
unless user_session.import_into_gb?
  @gbp_to_eur_exchange_rate = Api::MonetaryExchangeRate.for('GBP').exchange_rate.round(4)
end
```

I've not actually touched that here, though

alexdesi: oh, you didn't! sorry
... your code is clearer anyway

alexdesi (approved):
it looks good to me

PR #311, opened 2021-09-03
Standardise OptionCollection api

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-924

### What?

I have added/removed/altered:

- [x] Move to delegator over enumerable
- [x] Verify that wrapped methods wrap properly

### Why?

I am doing this because:

- OptionCollection wasn't returning an OptionCollection when calling reject
- OptionCollection needs to return an instance of itself everything certain array methods are called on the delegated subject
- Avoid regressions in behaviour

### DIFF
--- a/app/models/option_collection.rb
+++ b/app/models/option_collection.rb
@@ -1,28 +1,21 @@
-class OptionCollection
-  include Enumerable
+class OptionCollection < SimpleDelegator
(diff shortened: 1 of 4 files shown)


willfish:
Approved by @alexdesi

PR #313, opened 2021-09-15
Handle a partially answered journey with a user session prefill step

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-942

### What?

We need to handle prefilling of the user session such that the user can
be redirected to the correct location depending on the answers they've
already given on a different service.

The user session prefill step is responsible for receiving params that
have been answered (currently by the Single Trade Window project) and
redirecting the user to the appropriate step based on the established
routing logic for all of the different journeys of the duty calculator

The following routes are possible:

#### EU -> GB

`/duty-calculator/prefill?commodity_code=0702000007&country_of_origin=FR&import_date=2021-01-01&import_destination=UK`

#### NI -> GB

`/duty-calculator/prefill?commodity_code=0702000007&country_of_origin=XI&import_date=2021-01-01&import_destination=UK`

#### ROW -> GB

`/duty-calculator/prefill?commodity_code=0702000007&country_of_origin=NI&import_date=2021-01-01&import_destination=UK`

#### EU -> XI

`/duty-calculator/prefill?commodity_code=0702000007&country_of_origin=FI&import_date=2021-01-01&import_destination=XI`

#### ROW -> XI

`/duty-calculator/prefill?commodity_code=0702000007&country_of_origin=NI&import_date=2021-01-01&import_destination=XI`

#### GB -> XI
`/duty-calculator/prefill?commodity_code=0702000007&country_of_origin=GB&import_date=2021-01-01&import_destination=XI`

I have added/removed/altered:

- [x] Added a prefill route, controller and step to encapsulate the behaviour expected under different parameters being passed by the Single Trade Window team
- [x] Added specs to make sure all of the different routes are tested
- [x] Adds session builder for params method to the user session

### Why?

I am doing this because:

- This is a requirement to improve the user experience from the perspective of the Single Trade Window user
- The alternative would be to duplicate questions/answer twice and create a less seamless experience.


app/helpers/service_helper.rb
```diff
@@ -19,6 +19,10 @@ def tools_url
     service_url_for('/tools')
   end
 
+  def previous_service_url(commodity_code)
```
alexdesi, line 22: with previous, we mean the service who called trade tariff right?
that's not super clear, 
you have a better name? maybe `caller_service_url` ...

app/models/steps/prefill.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def next_step_path
+      return duty_path if user_session.ni_to_gb_route? || user_session.eu_to_ni_route?
+
+      return next_step_for_gb_to_ni if user_session.gb_to_ni_route?
+
+      return next_step_for_row_to_gb if user_session.row_to_gb_route?
+
+      next_step_for_row_to_ni
+    end
```
alexdesi, line 13: I am not a big fan of cascade IFs, it makes this hard to follow,
and actually, this is hard to follow 😄 for me
I've no idea how to make it clear at the moment, I'll thy to think something 🤔

app/models/steps/prefill.rb
```diff
@@ -0,0 +1,37 @@
...
+      return duty_path if user_session.zero_mfn_duty
+
+      trader_scheme_path
+    end
+
+    def next_step_for_row_to_ni
+      return interstitial_path if user_session.trade_defence
+      return trader_scheme_path unless user_session.zero_mfn_duty
+
+      user_session.commodity_source = 'uk'
```
alexdesi, line 28: this could create complications, 
this method is only to return the next step right? 
it should not change the commodity source (at least not inside this method)

alexdesi (commented):
Hi Will, I am impressed with how fast you made this 😄  I added some comments to clean things a bit.

app/models/user_session.rb
```diff
@@ -9,6 +11,47 @@ def initialize(session)
...
+    import_destination = params[:import_destination]
+    country_of_origin = params[:country_of_origin]
+    service = (import_destination == 'XI' ? 'xi' : 'uk')
+    other_country_of_origin = ''
+
+    # This reflects an idiosyncrasy in the xi_option form for country_of_origin and how we store the value on the session slightly differently. We'd need to refactor the way we access that if we want to remove this and just use country_of_origin as per the gb and non-row ni routes.
+    if import_destination == 'XI' && !Api::GeographicalArea.eu_member?(country_of_origin)
+      other_country_of_origin = country_of_origin
+      country_of_origin = 'OTHER'
+    end
```
alexdesi, line 24: this logic seems not to belongs here,
to start we could extract this logic to two private methods:

```
private 
other_country_of_origin(import_destination, eu_member)
country_of_origin(import_destination, eu_member)
```

to isolate it, so it'd be easy to move in the appropriate class.

spec/controllers/steps/prefill_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+        redirect_to: 'http://localhost/flibble',
+      }
+    end
+
+    before do
+      allow(UserSession).to receive(:build_from_params).and_call_original
+    end
+
+    it 'builds the session from params' do
+      response
```
alexdesi, line 22: what about renaming `response` with an "action"?
it sounds strange just to call "response"

possible name: call_show, perform_show, do_something, .... up to you

spec/helpers/service_helper_spec.rb
```diff
@@ -131,6 +133,33 @@
...
+    let(:user_session) { build(:user_session) }
+
+    context 'when redirect_to is set' do
+      let(:user_session) { build(:user_session, redirect_to: 'https://example.com/chieg') }
+
+      it { expect(helper.previous_service_url(commodity_code)).to eq('https://example.com/chieg') }
+    end
+
+    context 'when TRADE_TARIFF_FRONTEND_URL is set' do
+      it 'returns the dev trade tariff tools url' do
```
alexdesi, line 147: it is not clear from the test that it is actually set

app/controllers/steps/prefill_controller.rb
```diff
@@ -0,0 +1,38 @@
+module Steps
+  class PrefillController < BaseController
```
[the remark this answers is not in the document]
willfish: I think it's the context. The whole application is about generating a calculation by taking answers as part of a series of steps in a wizard. What else would you prefill in this single responsibility application?

Also, for reference: https://www.yourdictionary.com/prefill

app/controllers/steps/prefill_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    after_action :update_user_session
```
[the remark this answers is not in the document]
willfish: I don't think so. I trust the developer, here

app/controllers/steps/prefill_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      redirect_to @step.next_step_path
+    end
+
+    private
+
+    def user_session
+      @user_session ||= UserSession.build_from_params(session, user_session_params)
+    end
+
+    def user_session_params
```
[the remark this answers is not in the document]
willfish: It is private :)

app/helpers/service_helper.rb
willfish: The caller to me is the person making the current request, not the previous service they arrived from

spec/models/steps/prefill_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+
+        context 'when there is no zero mfn duty' do
+          let(:zero_mfn_duty) { false }
+
+          it { expect(step.next_step_path).to eq(trader_scheme_path) }
+        end
+      end
+    end
+  end
+end
```
alexdesi, line 105: these tests are so helpful! 👍

app/models/steps/prefill.rb
willfish: I don't think there is a way around it, sadly. The flow of the routing for the duty calculator is just complex :(

app/models/steps/prefill.rb
willfish: This is an idiom we've followed elsewhere, sadly. Possibly we could take the side effect out of the routing but we'd have to ask the routing questions twice

app/models/user_session.rb
willfish: This method is about orchestrating the variables that are required for building the user session from the parameters. We'd end up with lots of small methods where the method calls themselves are as big as the method body which also feels wrong to me.

We could isolate the user session builder into a service. Ultimately what it's building though is the user session so it feels appropriate that this goes here

willfish: I do think there's scope for clearing up the getters/setters on the user session, though. This could simplify the user session abstraction some what

spec/controllers/steps/prefill_controller_spec.rb
willfish: I think we could but we'd be going against a common pattern in the app. Maybe something for a separate refactor as it's much worse to use different patterns in the same app than getting one that you feel is right and another you feel is wrong or do a full refactor in the same PR and end up with unrelated changes.

app/controllers/steps/prefill_controller.rb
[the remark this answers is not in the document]
willfish: Maybe PrefillUserSessionController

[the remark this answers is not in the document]
willfish: It's a little weird because of the Verb/Noun nature of RPC/REST, too

spec/helpers/service_helper_spec.rb
willfish: I agree with this. I'll update it in both places that we validate this

app/controllers/steps/prefill_controller.rb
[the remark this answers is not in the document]
alexdesi: haha sorry 😄 I just missed that

app/controllers/steps/prefill_controller.rb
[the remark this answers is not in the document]
alexdesi: yeah, I think `PrefillUserSessionController` is a good name, that's basically what it is doing.
but if you are happy with the less explicit `PrefillController` no problem :)

app/controllers/steps/prefill_controller.rb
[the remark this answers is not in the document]
alexdesi: that's fine ( I just was thinking what's happen if in the worst-case scenario user_session is called without user_session_params, it probably will show up an error )

app/controllers/steps/prefill_controller.rb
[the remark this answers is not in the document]
willfish: Renamed the controller :)

alexdesi (commented):
✅

PR #317, opened 2021-09-15
Refactor user session

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Refactor the user session to test and implement answer and non-answer attributes in a standardised way

### Why?

I am doing this because:

- Extending the session model gets more and more involved and this dries it up significantly


app/models/concerns/user_session_attributes.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      define_method("#{attribute}=") do |value|
+        return session['answers'][attribute.to_s] = value if answer
+
+        session[attribute.to_s] = value
+      end
+    end
+
+    def dual_route_attribute(attribute)
```
alexdesi, line 19: Nice idea here!
the name was a bit strange at first, it's like route reminded me of the routing,
maybe we could be more explicit, maybe something like  `attribute_for_uk_and_xi`

alexdesi (approved):
I really like the idea and the implementation,
and nice cleaning!

just a comment about naming

PR #324, opened 2021-09-20
Fix gb to ni country of origin and add test

willfish:
### Jira link

QA Fixes

### What?

I have added/removed/altered:

- [x] Fixes broken gb to ni route

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -34,37 +34,47 @@ class UserSession
     @session['answers'][Steps::Excise.id] ||= {}
   end
 
-  def self.build_from_params(session, params)
-    import_destination = params[:import_destination]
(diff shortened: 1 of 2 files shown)


app/models/user_session.rb
```diff
@@ -34,37 +34,47 @@ def initialize(session)
...
+      Thread.current[:user_session]
+    end
+    
+    private
 
-  def self.get
-    Thread.current[:user_session]
+    def use_other_country_of_origin?(import_destination, country_of_origin)
+      import_destination == 'XI' &&
+        country_of_origin != 'GB' &&
```
willfish, line 75: This is the change

madviv (approved):

PR #329, opened 2021-09-27
HOTT-966 Content changes

alexdesi:
### What?
Content changes to the "When will the goods be imported" screen

### Why?
I am doing this because:
the new content is more readable

### Jira link
https://transformuk.atlassian.net/browse/HOTT-966


alexdesi:
Here is the page updated:
Screenshot 2021-09-27 at 14 51 19

jebw (approved):
LGTM

PR #334, opened 2021-10-07
Fixes monetary exchange rates

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1017

### What?

I have added/removed/altered:

- [x] Make as_of a global query param that pulls from the active session context import date value
- [x] Make monetary exchange rate be the rate at the time of the import
- [x] Adds build and build_collection specs to verify the global query
- [x] Adds specs for the monetary exchange rate changes

### Why?

I am doing this because:

- This is needed to make all requests to the api respect the import date specified by the user and to pull out the exchange rate that applies on the import date

### DIFF
--- a/spec/factories/api/monetary_exchange_rate.rb
+++ b/spec/factories/api/monetary_exchange_rate.rb
@@ -1,0 +1,11 @@
+FactoryBot.define do
+  sequence(:monetary_exchange_rate_id)
+
+  factory :monetary_exchange_rate, class: 'Api::MonetaryExchangeRate' do
+    id { generate(:monetary_exchange_rate_id) }
+    child_monetary_unit_code { 'GBP' }
+    exchange_rate { '0.7298' }
+    operation_date { nil }
+    validity_start_date { '2016-01-01T00:00:00.000Z' }
+  end
+end
--- a/app/models/api/monetary_exchange_rate.rb
+++ b/app/models/api/monetary_exchange_rate.rb
@@ -6,16 +6,32 @@ module Api
                :operation_date,
                :validity_start_date
 
-    def self.for(currency)
-      monetary_exchange_rates = build_collection(:xi).select do |monetary_exchange_rate|
-        monetary_exchange_rate.child_monetary_unit_code == currency.upcase
-      end
-
-      monetary_exchange_rates.max_by(&:validity_start_date)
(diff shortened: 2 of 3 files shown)


spec/factories/api/monetary_exchange_rate.rb
```diff
@@ -0,0 +1,11 @@
+FactoryBot.define do
+  sequence(:monetary_exchange_rate_id)
+
+  factory :monetary_exchange_rate, class: 'Api::MonetaryExchangeRate' do
+    id { generate(:monetary_exchange_rate_id) }
+    child_monetary_unit_code { 'GBP' }
```
alexdesi, line 6: `child_monetary_unit_code` does not sound a great naming 🤔 ... but I think it was already there

spec/models/api/base_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+    let(:id) { '0103921100' }
+    let(:query) { { foo: :bar } }
+
+    context 'when the service is uk' do
+      let(:service) { 'uk' }
+
+      it 'calls the uk client with the correct params' do
+        api_resource.build(service, id, query)
+
+        expect(Rails.application.config.http_client_uk).to have_received(:retrieve).with('commodities/0103921100.json', 'as_of' => Time.zone.today.iso8601, foo: :bar)
```
alexdesi, line 21: should we use symbol here?
`'as_of' => ...`

app/helpers/commodity_helper.rb
```diff
@@ -18,8 +17,9 @@ def xi_filtered_commodity
   def commodity
     commodity_source = user_session.commodity_source
     commodity_code = user_session.commodity_code
+    query = {}
 
-    commodity_context_service.call(commodity_source, commodity_code, default_query)
+    commodity_context_service.call(commodity_source, commodity_code, query)
```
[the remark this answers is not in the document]
willfish: I thought about this in terms of flagging to the user that they can supply a query but happy to change this

spec/factories/api/monetary_exchange_rate.rb
willfish: Yeah, this is what was chosen by the EU I believe - not by us.

PR #339, opened 2021-10-18
Fixes what it means to be on the deltas applicable route

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1043

### What?

I have added/removed/altered:

- [x] Extend the meaning of deltas applicable
- [x] Update trait to make tests pass with new meaning

### Why?

I am doing this because:

- This is required to correctly implement the Northern Ireland protocol


jebw (changes requested):
Can we get some tests covering the new conditions on deltas applicable?

PR #349, opened 2021-10-29
Adds resolved_measure_components as measure relationship

willfish:
### Jira link

HOTT-1009

### What?

I have added/removed/altered:

- [x] Added surfaced resolved measure components to the measure model

### Why?

I am doing this because:

- This is required to work with resolved components as part of duty calculations for meursing measures


alexdesi (approved):
good one!

PR #351, opened 2021-11-01
HOTT-1009: Adds step to take input for the meursing additional code

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1009

### What?

Screenshot from 2021-11-01 17-38-36

I have added/removed/altered:

- [x] Added step to take input for the meursing additional code
- [x] Make sure the step fits into the row to ni and gb to ni journeys
- [x] Updated certificate of origin step to go to interstitial page with a negative answer
- [x] Updated interstitial page to include the certificate of origin variation of the interstitial message

### Why?

I am doing this because:

- This is needed so we can resolve measures that have measure components with meursing duty expressions and then calculate duties for them

### DIFF
--- a/app/controllers/steps/meursing_additional_codes_controller.rb
+++ b/app/controllers/steps/meursing_additional_codes_controller.rb
@@ -1,0 +1,23 @@
+module Steps
+  class MeursingAdditionalCodesController < BaseController
+    def show
+      @step = Steps::MeursingAdditionalCode.new(meursing_additional_code_params)
+    end
+
+    def create
+      @step = Steps::MeursingAdditionalCode.new(meursing_additional_code_params)
+
+      validate(@step)
+    end
+
+    private
+
+    def meursing_additional_code_params
+      if params[:steps_meursing_additional_code].present?
+        params.require(:steps_meursing_additional_code).permit(:meursing_additional_code)
+      else
+        {}
+      end
+    end
+  end
+end
--- a/app/controllers/steps/meursing_additional_codes_controller.rb
+++ b/app/controllers/steps/meursing_additional_codes_controller.rb
@@ -13,11 +13,7 @@ module Steps
     private
 
     def meursing_additional_code_params
-      if params[:steps_meursing_additional_code].present?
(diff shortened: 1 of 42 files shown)


app/views/steps/interstitial/show.html.erb
```diff
@@ -3,7 +3,11 @@
 <span class="govuk-caption-xl">Calculate import duties</span>
 
 <% if user_session.gb_to_ni_route? %>
-  <%= render partial: 'steps/interstitial/shared/context', locals: { heading: t('interstitial.gb_to_ni.heading'), body: t('interstitial.gb_to_ni.body') } %>
+  <% if user_session.trade_defence? %>
```
willfish, line 6: I'm going to simplify this with an interstitial helper: https://transformuk.atlassian.net/browse/HOTT-1083.

A view shouldn't have this much logic in it IMO

app/controllers/steps/meursing_additional_codes_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    private
+
+    def meursing_additional_code_params
+      if params[:steps_meursing_additional_code].present?
+        params.require(:steps_meursing_additional_code).permit(:meursing_additional_code)
+      else
+        {}
+      end
```
alexdesi, line 20: I think you can use fetch here, something like:
`params.fetch(foo:, {}).permit(:bar, :baz)`

(found on https://stackoverflow.com/questions/17687506/how-to-make-an-optional-strong-parameters-key-but-filter-nested-params)

app/helpers/meursing_additional_code_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+    hint_text = t(
+      'meursing_additional_code.hint_text',
+      link: meursing_lookup_link,
+    )
+
+    sanitize(
+      hint_text,
+      tags: %w[p a],
+      attributes: %w[class target rel href],
+    )
```
alexdesi, line 12: I am not sure if this is used?

alexdesi (approved):
✅

PR #353, opened 2021-11-02
HOTT-1010: Reviewable Meursing additional code

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1010

### What?

Screenshot from 2021-11-02 12-00-39

I have added/removed/altered:

- [x] Added Meursing additional code reviewable answer to confirmation page

### Why?

I am doing this because:

- This is a business requirement


app/decorators/confirmation_decorator.rb
```diff
@@ -31,6 +32,7 @@ def path_for(key:)
     return additional_codes_path(applicable_measure_type_ids.first) if key == 'additional_code'
     return document_codes_path(document_codes_applicable_measure_type_ids.first) if key == 'document_code'
     return excise_path(applicable_excise_measure_type_ids.first) if key == 'excise'
+    return meursing_additional_codes_path if key == 'meursing_additional_code'
```
alexdesi, line 35: I agree with you that sometimes the multi-return pattern is handy, 
in this case I think, the `case/when` seems more readable (no returns and less duplication of ==)
```
case key
  when 'import_date'
    additional_codes_path(applicable_measure_type_ids.first)
  when 'document_code'
   document_codes_path(document_codes_applicable_measure_type_ids.first)
   ...
else
  send("#{key}_path")
end

```

alexdesi: up to you anyway, I am not strongly opinionated on this

alexdesi (approved):
It looks good ✅ 
just a minor comment.

PR #356, opened 2021-11-04
Adds resolved duty expression and resolved measure components

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Added missing fixture information to commodities after change in api

### Why?

I am doing this because:

- I want the fixtures to reflect the real responses


alexdesi (approved):
✅

PR #357, opened 2021-11-04
HOTT-1011: Calculate meursing duties

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1011

### What?

Screenshot from 2021-11-04 14-11-18

When a measure expresses meursing components and we have an answer from the user
about which code applies to their trade, we have enough information to
have the backend api surface the resolved measure components that we can
then use to execute duty calculations on the measure option that a
trader then take advantage of.

I've also updated the compound evaluator to express resolved meursing
duty expressions by concatenating the duty expression with a \<br\> tag
and the resolved duty expression coming back from the api response.

I have added/removed/altered:

- [x] Extend applicable measure components to include the resolved components
- [x] Update xi filtered commodities to include the meursing additional code so we retrieve the resolved meursing components
- [x] Use meursing components in the duty duty calculator over the standard components
- [x] Fill in test coverage for changed implemnetations

### Why?

- XI currently has measures with meursing components that we cannot currently calculate

### DIFF
--- a/app/controllers/steps/duty_controller.rb
+++ b/app/controllers/steps/duty_controller.rb
@@ -11,17 +11,17 @@ module Steps
 
     def duty_options
       return nil if user_session.no_duty_to_pay?
-      return DutyCalculator.new(filtered_commodity).options unless user_session.deltas_applicable?
+      return send("#{user_session.commodity_source}_options") unless user_session.deltas_applicable?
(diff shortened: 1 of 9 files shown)


jebw (approved):
Looks good, nice feature!

PR #361, opened 2021-11-12
HOTT-1111: Make attribute api more robust to new attributes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1111

### What?

I have added/removed/altered:

- [x] Move to a more robust attributes api

### Why?

I am doing this because:

- This is required to support new attributes that aren't defined on the models being returned by the backend not causing exceptions

### DIFF
--- a/app/models/api/base.rb
+++ b/app/models/api/base.rb
@@ -1,13 +1,29 @@
 module Api
   class Base
+    attr_accessor :attributes
+
+    delegate :user_session, to: :class
+    delegate :as_json, to: :attributes
+
+    def initialize(attributes = {})
+      @attributes = ActiveSupport::HashWithIndifferentAccess.new(attributes.fetch('attributes', attributes))
+    end
+
     class << self
       def inherited(child)
         super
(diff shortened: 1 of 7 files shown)


app/models/api/base.rb
```diff
@@ -1,15 +1,31 @@
 module Api
   class Base
+    attr_accessor :attributes
+
+    delegate :user_session, to: :class
```
willfish, line 5: This is hoisted so that all delegations are at the top and is not new

spec/models/api/base_spec.rb
```diff
@@ -6,6 +6,21 @@
     allow(Rails.application.config.http_client_xi).to receive(:retrieve).and_call_original
   end
 
+  describe '#initialize' do
+    context 'when initialized with unknown attributes' do
```
willfish, line 10: Validates that we can instantiate a model without knowing about the attributes explicitly

jebw (approved):

PR #364, opened 2021-11-18
HOTT-1119: Separate browse and search urls

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1119

### What?

Screenshot from 2021-11-18 12-01-30

I have added/removed/altered:

- [x] Added separate nav items for browse and search when using the updated navigation
- [x] Refactored and improved common tests of service urls using shared examples

### Why?

I am doing this because:

- This is needed to make sure that the duty calculator reflects what's currently on the other environments


alexdesi (approved):
looks good!

PR #365, opened 2021-11-18
HOTT-1130: Use UK measures for excise measure units

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1130

### What?

I have added/removed/altered:

- [x] Updates all steps to use the ApplicableMeasureUnitMerger where relevant
- [x] Adds specs to avoid regressions
- [x] Refactors some specs for consistency

### Why?

I am doing this because:

- The UK is the source of truth for excise measure units and we should be using the ApplicableMeasureUnitMerger when determining measure units to know these


willfish:
Approved verbally by Madhu

PR #369, opened 2021-11-29
HOTT-1109: Extract complex document code extraction into own service

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1109

### What?

I have added/removed/altered:

- [x] Extracted ApplicableDocumentCodesService
- [x] Added ApplicableDocumentCodesService coverage

### Why?

I am doing this because:

- This is an improvement in itself but it makes fixing the linked ticket issue easier

### DIFF
--- a/app/helpers/commodity_helper.rb
+++ b/app/helpers/commodity_helper.rb
@@ -53,15 +53,7 @@ module CommodityHelper
   end
 
   def applicable_document_codes
-    @applicable_document_codes ||=
-      {}.tap do |applicable_document_codes|
-        if user_session.deltas_applicable?
-          applicable_document_codes['uk'] = surface_document_codes(commodity: uk_filtered_commodity)
-          applicable_document_codes['xi'] = surface_document_codes(commodity: xi_filtered_commodity)
-        else
-          applicable_document_codes[user_session.commodity_source] = surface_document_codes
-        end
-      end
+    @applicable_document_codes ||= ApplicableDocumentCodesService.new.call
   end
 
   def applicable_document_codes?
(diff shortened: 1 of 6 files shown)


app/helpers/commodity_helper.rb
```diff
@@ -53,15 +53,7 @@ def applicable_excise_additional_codes
...
-    @applicable_document_codes ||=
-      {}.tap do |applicable_document_codes|
-        if user_session.deltas_applicable?
-          applicable_document_codes['uk'] = surface_document_codes(commodity: uk_filtered_commodity)
-          applicable_document_codes['xi'] = surface_document_codes(commodity: xi_filtered_commodity)
-        else
-          applicable_document_codes[user_session.commodity_source] = surface_document_codes
-        end
-      end
+    @applicable_document_codes ||= ApplicableDocumentCodesService.new.call
```
jebw, line 56: You could inject the dependency to avoid having a hard dependency on `UserSession.get` - it would avoid needing to stub the `UserSession.get` method in specs?

`@applicable_document_codes ||= ApplicableDocumentCodesService.new(UserSession.get).call`

app/services/applicable_document_codes_service.rb
```diff
@@ -0,0 +1,47 @@
+class ApplicableDocumentCodesService
+  include CommodityHelper
+
+  def initialize
```
jebw, line 4: tied in with above

```ruby
def initialize(user_session)
  @user_session = user_session
end
```

jebw (approved):
LGTM

app/helpers/commodity_helper.rb
willfish: Feels a bit weird injecting a global :thinking:

willfish: I'll have a look in the follow on PR since I'm making changes to these tests

PR #372, opened 2021-11-30
HOTT-1109: Fix bug in document codes presentation

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1109

### What?

I have added/removed/altered:

- [x] Altered the document codes service to make sure document codes for a given measure type id are unique, ordered and with the last option always being 'None'
- [x] Updated the specs to reflect this

**Before**

Screenshot from 2021-11-30 11-23-15

**After**

Screenshot from 2021-11-30 11-28-01

### Why?

I am doing this because:

- This commodity for Norway https://staging.trade-tariff.service.gov.uk/xi/commodities/1905320500?country=NO#import
shows up two None options that aren't at the end of the list and presents incorrectly in the UI

### DIFF
--- a/app/services/applicable_document_codes_service.rb
+++ b/app/services/applicable_document_codes_service.rb
@@ -6,9 +6,13 @@ class ApplicableDocumentCodesService
   end
 
   def call
-    return delta_measure_condition_documents if user_session.deltas_applicable?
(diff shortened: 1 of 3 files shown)


willfish:
Approved by @madviv

PR #373, opened 2021-12-01
HOTT-1143: Upgrade to supported sentry-rails

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1143

### What?

I have added/removed/altered:

- [x] Move from the old sentry-raven gem to the sentry-rails gem

### Why?

I am doing this because:

- The old one is out of support even for bug fixes

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'logstash-event'
 gem 'newrelic_rpm'
 gem 'puma'
 gem 'railties', '~> 6'
-gem 'sentry-raven'
+gem 'sentry-rails'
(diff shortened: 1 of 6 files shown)


willfish:
The original implementation of the background worker is:

```ruby
module Sentry
  class BackgroundWorker
    def _perform(&block)
      # make sure the background worker returns AR connection if it accidentally acquire one during serialization
      ActiveRecord::Base.connection_pool.with_connection do
        block.call
      end
    end
  end
end
```

All I'm doing is removing the hard dependency on ActiveRecord since we don't use it in any of our projects

jebw (approved):

PR #375, opened 2021-12-06
HOTT-1182: Handle compound measure units

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1182

### What?

When ASVX measure units are encountered in the backend they are replaced
by two measure units for volume and alcohol percentage.

We need these to be handled in the duty calculator slightly differently.

This maybe extended later if we're calculating other measure units
differently.

I have added/removed/altered:

- [x] Added a compound measure unit evaluator to handle compound measure units
- [x] Extended evaluator factory specs to include the CompoundMeasureUnit evaluator in the single and compound component case
- [x] Simplified current evaluator factory specs with a shared example

### Why?

I am doing this because:

- We were doing measure unit calculations on ASVX incorrectly

### DIFF
--- a/app/services/expression_evaluators/compound_measure_unit.rb
+++ b/app/services/expression_evaluators/compound_measure_unit.rb
@@ -1,0 +1,76 @@
+module ExpressionEvaluators
+  class CompoundMeasureUnit < ExpressionEvaluators::Base
+    include MeasureUnitPresentable
+
+    def call
+      {
+        calculation: calculation_duty_expression,
+        value: value,
+        formatted_value: number_to_currency(value),
+      }
+    end
+
+    private
+
+    def calculation_duty_expression
+      expression =
+        if measure_condition.present?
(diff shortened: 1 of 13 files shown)


app/services/expression_evaluators/compound_measure_unit.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    def presented_volume_unit
+      @presented_volume_unit ||= {
+        answer: user_session.measure_amount[Api::BaseComponent::VOLUME_UNIT.downcase],
+        unit: volume_unit['unit'],
+      }
+    end
+
+    def measure_condition
+      return nil if component.is_a?(Api::MeasureComponent)
```
alexdesi, line 63: is not this already checked externally?

alexdesi (approved):
it seems correct to me, 
but consider that I am a not duty-calculator expert 😄

PR #376, opened 2021-12-07
Tidies

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes some duplication between two different expression evaluators
- [x] Stops checking the type of the component to pull out the condition
- [x] Removes iteration by fetching the measure condition id from the
    measure condition component split version of its id
- [x] Adds coverage for these changes

### Why?

I am doing this because:

- This makes the code more readable and easier to work with by being more consistent with ruby/object-oriented idioms


alexdesi (approved):
💚 nice and sweet :)

PR #377, opened 2021-12-07
HOTT-1140: Revert to default govuk frontend width

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1140

### What?

See related PR for the frontend: 

I have added/removed/altered:

- [x] Revert to bundled default width specified by govuk-frontend

### Why?

I am doing this because:

- This is to be aligned with a recent change in the frontend project https://github.com/trade-tariff/trade-tariff-frontend/pull/450

### DIFF
--- a/app/webpacker/styles/application.scss
+++ b/app/webpacker/styles/application.scss
@@ -9,7 +9,6 @@
 $govuk-font-url-function: frontend-font-url;
 $govuk-image-url-function: frontend-image-url;
 $govuk-global-styles: true;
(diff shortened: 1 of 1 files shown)


jebw (approved):
LGTM

FYI: I've not yet merged my PR - I'm aware we're potentially going to need some quick turn around releases in the next couple of weeks so was waiting until the Review branches stuff was finished so Matt could review it on there before merging (and potentially immediately releasing)

PR #405, opened 2022-01-25
HOTT-1295: Prep for ruby 3.1.0

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1295

### What?

I have added/removed/altered:

- [x] Updates rails framework gems to 7.x
- [x] Unpins the formbuilder to a specific version

### Why?

I am doing this because:

- We need 7.x in order for us to support 3.1.0
- We need the formbuilder to be updated


Gemfile
```diff
@@ -2,19 +2,19 @@ source 'https://rubygems.org'
...
-gem 'actionpack', '~> 6'
-gem 'actionview', '~> 6'
-gem 'activemodel', '~> 6'
-gem 'activerecord', '~> 6'
-gem 'activesupport', '~> 6'
+gem 'actionpack'
+gem 'actionview'
+gem 'activemodel'
+gem 'activerecord'
+gem 'activesupport'
```
alexdesi, line 9: I guess we should specify 
`, '~> 6'`
in here.

willfish: Not sure :shrug:

willfish: Given we have a lock file and updates are very much under our control I don't see the advantage

alexdesi (approved):

Gemfile
willfish: Spoken about this - basically it reduces headaches where we have to manage updates that push us to rails 8 before we're ready

PR #423, opened 2022-02-14
Adds smoke test dev and staging step to duty calculator

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Added smoke tests to the ci workflows

### Why?

I am doing this because:

- To avoid regressions in the govuk-frontend upgrade that would have been caught by the smoke tests


jebw (approved):
LGTM :+1:

PR #426, opened 2022-02-17
HOTT-1295: Move to rails 7 configuration defaults

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Move to rails 7 configuration defaults

### Why?

I am doing this because:

- This is part of a 3 step upgrade to rails

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,19 +1,67 @@
+require "active_support/core_ext/integer/time"
+
 Rails.application.configure do
+  # Settings specified here will take precedence over those in config/application.rb.
+
+  # Code is not reloaded between requests.
   config.cache_classes = true
 
+  # Eager load code on boot. This eager loads most of Rails and
+  # your application in memory, allowing both threaded web servers
+  # and those relying on copy on write to perform better.
+  # Rake tasks automatically ignore this option for performance.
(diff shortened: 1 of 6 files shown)


config/environments/production.rb
```diff
@@ -1,19 +1,68 @@
...
+  # Specifies the header that your server uses for sending files.
+  # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
+  # config.action_dispatch.x_sendfile_header = "X-Accel-Redirect" # for NGINX
+
+  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
+  # config.force_ssl = true
 
+  # Include generic and useful information about system operation, but avoid logging too much
+  # information to avoid inadvertent exposure of personally identifiable information (PII).
+  config.log_level = :info
```
willfish, line 39: Production should definitely not be dumping debug log lines IMO

jebw: Yeh, probably right

jebw (approved):
LGTM - but the `service_url_for` could probably be using `URI.join` instead of `File.join` - kinda unrelated but looked into the logic of sections_url if we were allowing hosts

PR #427, opened 2022-02-17
Fix additional code labels for XI and UK.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1327

### What?
Fix additional code labels for XI and UK.

### DIFF
--- a/app/views/steps/additional_codes/show.html.erb
+++ b/app/views/steps/additional_codes/show.html.erb
@@ -6,11 +6,11 @@
   <h1 class="govuk-heading-xl">Describe your goods in more detail</h1>
 
   <% if @step.applicable_additional_codes.dig('uk', @step.measure_type_id).present? %>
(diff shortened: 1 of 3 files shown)


alexdesi:
Before:
Screenshot 2022-02-17 at 15 49 27

After:
Screenshot 2022-02-17 at 15 45 04

jebw (approved):
LGTM

willfish (dismissed):
We need to hide these

willfish:
We should add the same to the document code step if this is what @mattlavis-transform wants for all of the steps with missing labels

alexdesi:
@willfish I think you are right for the radio buttons, which already have a label for each item.
In that case I fixed the form and removed the explicit H1.

For the selectbox I added the label as discussed (all the other fields in other pages have already a label).

PR #436, opened 2022-03-01
Upgrade webpacker

jebw:
### Jira link

HOTT-????

### What?

I have added/removed/altered:

- [x] Dropped the unneeded dev dependency on webpack
- [x] Upgraded all npm depencies

### Why?

I am doing this because:

- We have various npm related security warnings


willfish (approved):
image

PR #438, opened 2022-03-11
HOTT-1402 Use tagged releases

jebw:
### Jira link

[HOTT-1402](https://transformuk.atlassian.net/browse/HOTT-1402)

### What?

I have added/removed/altered:

- [x] create a release + git tag in GitHub
- [x] tag the release in DockerHub
- [x] notify the slack channel when a release happens
- [x] allow re-releasing older releases 

### Why?

I am doing this because:

- We want to be able to redeploy releases without rebuilding
- We want to build a record of what we have released and when
- We want to notify the team when changes are released, and what was changed

- Makes changes to the deployment process

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -248,7 +248,7 @@ jobs:
           app_domain_prefix: "dev"
       - sentry-release
 
-  deploy_staging:
+  deploy_main_to_staging:
     docker:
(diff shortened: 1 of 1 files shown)


willfish (changes requested):
Let's chat about this at some point

willfish (approved):

PR #453, opened 2022-03-31
Bump UKTT to handle join entities

jebw:
### Jira link

HOTT-????

### What?

I have added/removed/altered:

- [x] Bumped the version of the UKTT gem

### Why?

I am doing this because:

- the new version handles join entities in JSONAPI correctly

- New version of UKTT now raises a custom error on parsing problems. This _shouldn't_ make a difference since there's no custom code for parsing errors in DC anyway AFAICT

### Testing done

- enabled Permutations API, verified that running main on Dev failed smoke tests
- deployed this PR, verified smoke tests passed


alexdesi (approved):
👍

PR #454, opened 2022-03-31
HOTT-1455: Removes unused method

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1455

### What?

I have added/removed/altered:

- [x] Removed unused method and spec

### Why?

I am doing this because:

- All code is overhead


jebw (approved):
LGTM

can we change the title to be what got changed (ie dropped unused xxxxx method) - this is what shows up in release notes?

PR #456, opened 2022-04-01
HOTT-1455: Adds support for a stopping page

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1455

### What?

Screenshot from 2022-04-01 14-39-39

For some measures that have a stopping negative measure condition we
need to halt all possibility of duty calculation and suggest actions to
the user.

I have added/removed/altered:

- [x] Added support for the stopping step and coverage
- [x] Removed complex factory initializer on measure

### Why?

I am doing this because:

- This is required to halt indicate to traders and HMRC staff that trading is stopped when stopping conditions are met


spec/factories/user_session.rb
```diff
@@ -243,4 +243,26 @@
...
+        },
+      }
+    end
+  end
+
+  trait :with_a_single_stopping_condition_document_answer do
+    document_code do
+      {
+        'uk' => {
+          '103' => 'None',
```
willfish, line 262: Consider removing this

spec/factories/user_session.rb
```diff
@@ -243,4 +243,26 @@
...
+          '105' => 'None',
+        },
+      }
+    end
+  end
+
+  trait :with_a_single_stopping_condition_document_answer do
```
willfish, line 258: Rename this to reflect one answer being positive

jebw (approved):
LGTM

PR #468, opened 2022-04-14
HOTT-1510: Ruby 3.1.2

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1510

### What?

I have added/removed/altered:

- [x] Bumps to ruby 3.1.2

### Why?

I am doing this because:

- This patch fixes a couple of security vulnerabilities
- https://github.com/ruby/ruby/compare/v3_1_1...v3_1_2
- https://www.ruby-lang.org/en/news/2022/04/12/ruby-3-1-2-released/


.circleci/config.yml
```diff
@@ -161,7 +161,7 @@ jobs:
 
   checking:
     docker:
-      - image: 'cimg/ruby:3.1.1-node'
```
jebw, line 164: I think this should reference `cimg/ruby:3.1-node` - ie _not_ specify the patch level - that way we automatically benefit from patch level updates without doing any accidental ruby version bumps

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build compilation image
-FROM ruby:3.1.1-alpine3.15 as builder
```
jebw, line 2: I think this should reference `ruby:3.1-alpine3.15` - ie _not_ specify the patch level - that way we automatically benefit from patch level updates without doing any accidental ruby version bumps

Dockerfile
```diff
@@ -40,7 +40,7 @@ RUN rm -rf node_modules log tmp && \
       find /usr/local/bundle/gems -name "*.html" -delete
 
 # Build runtime image
-FROM ruby:3.1.1-alpine3.15 as production
```
jebw, line 43: I think this should reference ruby:3.1-alpine3.15 - ie not specify the patch level - that way we automatically benefit from patch level updates without doing any accidental ruby version bumps

jebw (commented):
I think we should be not specifying the patch level - I cannot see us wanting to run against an out of date patch release?

jebw (approved):

PR #470, opened 2022-04-19
HOTT-1481: Normalise user inputs with multipliers

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1481

### What?

When the user enters a unit who's input type is different from the type
used in the calculator we need to use the multiplier to coerce the input answer
to a normalised value (e.g. convert liters back into hectoliters). This enables
the calculator to accurately fill in the unit answers required by the current component.

We've also migrated away from a dependence on massive fixture files with
difficult-to-configure attributes and are now instead using a commodity
factory.

When the multiplier isn't present we do no coercion of input answers.

Finally, we introduce a whole bunch of orchestrating factory traits that
make simulation of various conditions in the evaluators more straightforward

I have added/removed/altered:

- [x] Added application of multipliers in measure unit, compound measure unit evaluators
- [x] Added coverage for missing/present multiplier behaviour
- [x] Altered existing specs to use a commodity factory
- [x] Added codification of measure unit presentable behaviour in shared examples
- [x] Added a bunch of traits for easier control of different fixture inputs driving different behaviours

### Why?

I am doing this because:

- This is all required because we're simplifying the interface of the duty
calculator for the user such that they have a limited number of more
commonly known unit types to enter as part of the duty calculator
process.

### NB

I've actually not supported normalised retail price measure units since this is just a value in pounds that is unlikely to ever
require normalising with a multiplier.

### DIFF
--- a/spec/services/expression_evaluators/retail_price_spec.rb
+++ b/spec/services/expression_evaluators/retail_price_spec.rb
@@ -9,9 +9,19 @@ RSpec.describe ExpressionEvaluators::RetailPrice, :user_session do
   let(:measure_component) { measure.measure_components.first }
 
   let(:commodity) do
-    measure = attributes_for(:measure, :excise, measure_components: [attributes_for(:measure_component, :with_retail_price_measure_units)])
-
-    build(:commodity, :with_retail_price_measure_units, import_measures: [measure])
+    build(
+      :commodity,
+      :with_retail_price_measure_units,
+      import_measures: [
+        attributes_for(
+          :measure,
+          :excise,
+          measure_components: [
+            attributes_for(:measure_component, :with_retail_price_measure_units),
+          ]
+        )
+      ]
+    )
   end
 
   let(:expected_evaluation) do
--- a/spec/services/expression_evaluators/retail_price_spec.rb
+++ b/spec/services/expression_evaluators/retail_price_spec.rb
@@ -1,9 +1,18 @@
 RSpec.describe ExpressionEvaluators::RetailPrice, :user_session do
   subject(:evaluator) { described_class.new(measure, measure_component) }
 
-  let(:measure) { build(:measure, :excise) }
-  let(:measure_component) { build(:measure_component, :with_retail_price_measure_units) }
-  let(:user_session) { build(:user_session, :with_retail_price_measure_amount, commodity_code: '0103921100') }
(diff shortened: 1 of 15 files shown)


spec/services/expression_evaluators/retail_price_spec.rb
```diff
@@ -1,9 +1,18 @@
...
-  let(:user_session) { build(:user_session, :with_retail_price_measure_amount, commodity_code: '0103921100') }
+  include_context 'with a fake commodity'
+
+  let(:user_session) { build(:user_session, :with_retail_price_measure_amount, commodity_code: commodity.code) }
+
+  let(:measure) { commodity.import_measures.first }
+  let(:measure_component) { measure.measure_components.first }
+
+  let(:commodity) do
+    measure = attributes_for(:measure, :excise, measure_components: [attributes_for(:measure_component, :with_retail_price_measure_units)])
```
alexdesi, line 12: this assignment could be problematic, 
we have already a let names `measure`,
do you want to use the `let` or the `var` ?

spec/support/shared_examples/a_measure_unit_presentable.rb
```diff
@@ -0,0 +1,13 @@
+RSpec.shared_examples_for 'a measure unit presentable' do
+  describe '#presented_unit' do
+    it 'returns the properly presented unit' do
+      expect(evaluator.presented_unit).to eq(expected_presented_unit) if expected_presented_unit
```
alexdesi, line 4: conditional expectation? 🤔 
I guess we should use a context here

alexdesi (approved):
really good!
just couple of comments on tests
✅

spec/support/shared_examples/a_measure_unit_presentable.rb
```diff
@@ -0,0 +1,13 @@
...
+    end
+  end
+
+  describe '#applicable_unit' do
+    it 'returns the merged units from the ApplicableMeasureUnitMerger service' do
+      expect(evaluator.applicable_unit).to eq(expected_applicable_unit)
+    end
+  end
+end
```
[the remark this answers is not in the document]
willfish: The evaluator is the subject :). I'm treating it as a parameter, actually. I could probably pull it explicitly into the shared example since the api is the same over and over for evaluators

PR #474, opened 2022-04-27
HOTT-1540: Deduplicate measurement units

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1540

**Before**

Screenshot from 2022-04-27 11-59-16

**After**

Screenshot from 2022-04-26 12-33-18
d3.png)

### What?

I have added/removed/altered:

- [x] Adds unit answer service and specs
- [x] Adds deduplication to the applicable measure unit merger
- [x] Consume answer from answer service
- [x] Removes unused fixture file
- [x] Fixes issue where the new unit presents on the duty screen when the original unit should be shown

### Why?

I am doing this because:

- This is required to improve the user experience/reduce confusion


app/decorators/confirmation_decorator.rb
```diff
@@ -76,13 +76,17 @@ def format_measure_amount(value, _key)
...
+      applicable_unit = applicable_measure_units[measure_unit_key.upcase]
+      has_coerced_unit = applicable_unit['coerced_measurement_unit_code'].present?
+      abbreviation = applicable_unit['abbreviation']
+      unit = applicable_unit['unit']
 
       if measure_unit_key.upcase == Api::BaseComponent::RETAIL_PRICE_UNIT
         "<span title='#{abbreviation}'>#{number_to_currency(answer, unit:)}</span>"
-      else
+      elsif !has_coerced_unit
         "<span title='#{abbreviation}'>#{answer} #{unit}</span>"
```
jebw, line 87: suggestion

`helpers.tag.span(%(#{answer} {unit}), title: abbreviation)` - this will then handle any escaping of special chars properly

might be `helper.` - can't remember

app/services/applicable_measure_unit_merger.rb
```diff
@@ -14,29 +18,35 @@ def call
   private
 
   def delta_measure_units
-    clean_unhandled_measure_units do
-      uk_filtered_commodity.applicable_measure_units.merge(xi_filtered_commodity.applicable_measure_units)
-    end
+    applicable_units =  uk_filtered_commodity.applicable_measure_units.merge(xi_filtered_commodity.applicable_measure_units)
```
jebw, line 21: thiis feels a bit hard to follow, two long names with `.merge` tucked away in the middle - how about

```ruby
uk_units = uk_filtered_commodity.applicable_measure_units
xi_units  = xi_filtered_commodity.applicable_measure_units
applicable_units = uk_units.merge(xi_units)

clean_and_dedupe_measure_units applicable_units
```

app/services/applicable_measure_unit_merger.rb
```diff
@@ -14,29 +18,35 @@ def call
...
+    applicable_units = uk_filtered_commodity.applicable_measure_units
+
+    clean_and_dedup_measure_units(applicable_units)
   end
 
   def xi_measure_units
-    clean_unhandled_measure_units do
-      xi_filtered_commodity.applicable_measure_units.merge(uk_filtered_commodity.applicable_excise_measure_units)
-    end
+    applicable_units = xi_filtered_commodity.applicable_measure_units.merge(uk_filtered_commodity.applicable_excise_measure_units)
```
jebw, line 33: Similar suggestion to `delta_measure_units` ?

app/services/applicable_measure_unit_merger.rb
```diff
@@ -14,29 +18,35 @@ def call
...
+    clean_and_dedup_measure_units(applicable_units)
   end
 
-  def clean_unhandled_measure_units
-    yield.tap do |units|
-      UNHANDLED_MEASURE_UNITS.each do |unhandled_unit|
-        units.delete(unhandled_unit)
+  def clean_and_dedup_measure_units(applicable_units)
+    if @dedupe
+      applicable_units = applicable_units.each_with_object({}) do |(measurement_unit_code, measurement_unit_values), acc|
```
jebw, line 40: `acc` isn't particular descriptive? `all_units` maybe ? 

 if line length is a concern then prefixing the `measurement_unit_` could probably be shortened to just `unit_` or even dropping the prefix entirely since the context is apparent from the method name ?

app/services/applicable_measure_unit_merger.rb
```diff
@@ -14,29 +18,35 @@ def call
...
+        acc[measurement_unit_values['coerced_measurement_unit_code'].presence || measurement_unit_code] = measurement_unit_values
       end
     end
+
+    UNHANDLED_MEASURE_UNITS.each do |unhandled_unit|
```
jebw, line 45: `applicable_units.without(UNHANDLED_MEASURE_UNITS)` ?

jebw (approved):
LGTM - I've included a couple of suggestions around syntax

app/decorators/confirmation_decorator.rb
willfish: Awesome, thanks :+1:

app/services/applicable_measure_unit_merger.rb
willfish: Yeah, happy to help with readability! I came back to this and had to remind myself what was going on so appreciate the suggestion

app/services/applicable_measure_unit_merger.rb
willfish: acc/accumulator is something that's super common when accumulating into an enumerable. we've definitely been here before and I'm happy to just go with what is easy to understand for other people

app/services/applicable_measure_unit_merger.rb
willfish: I think except would do it, yeah.

app/services/concerns/measure_unit_presentable.rb
```diff
@@ -13,9 +14,15 @@ def applicable_unit
...
+    key = unit['coerced_measurement_unit_code'].presence || "#{unit['measurement_unit_code']}#{unit['measurement_unit_qualifier_code']}"
+
+    user_session.measure_amount[key.downcase]
+  end
+
   private
 
   def applicable_units
-    @applicable_units ||= ApplicableMeasureUnitMerger.new.call
+    @applicable_units ||= ApplicableMeasureUnitMerger.new(dedupe: false).call
```
alexdesi, line 26: I think it's better to be more explicit:
`remove_duplication: true/false` , or `no_duplication`, or `without_duplicate`, ...
dedupe is not a word, isn't it?

alexdesi: Screenshot 2022-04-28 at 12 06 37

willfish:
Smoke tests are failing because the units have changed from dtn and hlt to kgm and ltr, respectively

app/services/concerns/measure_unit_presentable.rb
willfish: Spoke separately

PR #479, opened 2022-05-03
HOTT-1552 Exclude multiplier from duty calculation

jebw:
On measure units

### Jira link

[HOTT-1552](https://transformuk.atlassian.net/browse/HOTT-1552)

### What?

I have added/removed/altered:

- [ ] Changed display of duty expression to exclude the quantity string

### Why?

I am doing this because:

- The string is confusing for the user


alexdesi (approved):
It looks great ✅

PR #525, opened 2022-08-25
HOTT-1909: Handle compound sucrose measurement units

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1909

### What?

We evaluate (read calculate based on inputs from different places) the expressions of measure components of various kinds. Sometimes these components are expressed as a simple percentage (ad valorem) where the inputs come from the backend measure component itself (the duty amount) and sometimes they take account of the input measurement units from the user (stored in the user session).

We've recently discovered that the DTNZ type of measurement unit measure component needed to be special cased/evaluated using slightly different rules to a normal/simple measurement unit and that it actually required the input answers of two questions: What is the weight, and what is the percent sucrose content?

After updating the backend to make the relevant changes to what questions need to be answered when we encounter these measurement units, we now have to update the duty calculator to reflect the change to how these unit components get calculated (which is the result of plumbing the sucrose evaluator together when we encounter these measure components).

I have added/removed/altered:

- [x] Tidy dotenv files
- [x] Bump uktt to support basic auth locally
- [x] Adds enum for base component unit
- [x] Plumb sucrose evaluator in measure factory
- [x] Adds sucrose evaluator for DTNZ

**BEFORE**

image

**AFTER**

image

### Why?

I am doing this because:

- We're currently doing the wrong calculations for DTNZ and are treating DTNZ as a simple coerced KGM unit.

### DIFF
--- a/app/services/expression_evaluators/sucrose_measure_unit.rb
+++ b/app/services/expression_evaluators/sucrose_measure_unit.rb
@@ -4,7 +4,7 @@ module ExpressionEvaluators
 
     def call
       {
-        calculation: calculation_duty_expression,
+        calculation: sanitized_duty_expression,
         value:,
         formatted_value: number_to_currency(value),
       }
@@ -12,19 +12,14 @@ module ExpressionEvaluators
 
     private
 
-    def calculation_duty_expression
-      expression =
-        if measure_condition.present?
-          measure_condition.duty_expression
-        else
-          measure.duty_expression.formatted_base
-        end
+    def sanitized_duty_expression
+      expression = measure_condition&.duty_expression || measure.duty_expression.formatted_base
 
       sanitize(expression, tags: %w[span abbr], attributes: %w[title])
     end
 
     def value
-      candidate_value = component.duty_amount * decitonne_quantity * sucrose_quantity
+      candidate_value = component.duty_amount * quantity_in_decitonnes * sucrose_quantity
 
       if component.euros?
         candidate_value * euro_exchange_rate
@@ -33,7 +28,7 @@ module ExpressionEvaluators
       end
     end
 
-    def decitonne_quantity
+    def quantity_in_decitonnes
       presented_decitonne_unit[:answer].to_f * presented_decitonne_unit[:multiplier].to_f
     end
 
--- a/app/services/expression_evaluators/sucrose_measure_unit.rb
+++ b/app/services/expression_evaluators/sucrose_measure_unit.rb
@@ -1,0 +1,71 @@
+module ExpressionEvaluators
+  class SucroseMeasureUnit < ExpressionEvaluators::Base
+    include MeasureUnitPresentable
+
+    def call
+      {
+        calculation: calculation_duty_expression,
+        value:,
+        formatted_value: number_to_currency(value),
+      }
+    end
+
+    private
+
+    def calculation_duty_expression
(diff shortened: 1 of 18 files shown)


app/services/expression_evaluators/sucrose_measure_unit.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    private
+
+    def calculation_duty_expression
+      expression =
+        if measure_condition.present?
+          measure_condition.duty_expression
+        else
+          measure.duty_expression.formatted_base
+        end
```
alexdesi, line 21: alternative:
`expression = measure_condition&.duty_expression || measure.duty_expression.formatted_base`

app/services/expression_evaluators/sucrose_measure_unit.rb
```diff
@@ -0,0 +1,71 @@
...
+      {
+        calculation: calculation_duty_expression,
+        value:,
+        formatted_value: number_to_currency(value),
+      }
+    end
+
+    private
+
+    def calculation_duty_expression
```
alexdesi, line 15: what about renaming this to
`def duty_expression`

app/services/expression_evaluators/sucrose_measure_unit.rb
```diff
@@ -0,0 +1,71 @@
...
+      candidate_value = component.duty_amount * decitonne_quantity * sucrose_quantity
+
+      if component.euros?
+        candidate_value * euro_exchange_rate
+      else
+        candidate_value
+      end
+    end
+
+    def decitonne_quantity
```
alexdesi, line 36: `quantity_in_decitonne` ?

spec/models/api/base_component_spec.rb
```diff
@@ -118,6 +118,29 @@
     end
   end
 
+  describe '#sucrose?' do
+    subject(:component) do
+      described_class.new(
+        'measurement_unit_code' => measurement_unit_code,
+        'measurement_unit_qualifier_code' => measurement_unit_qualifier_code,
```
alexdesi, line 125: does not work with symbols:
`measurement_unit_code: measurement_unit_code,
...
`

spec/models/api/base_component_spec.rb
```diff
@@ -118,6 +118,29 @@
...
+      )
+    end
+
+    context 'when the unit is an sucrose volume unit' do
```
alexdesi, line 129: when the unit is a~n~ sucrose volume unit

spec/models/api/base_component_spec.rb
```diff
@@ -118,6 +118,29 @@
...
+      let(:measurement_unit_code) { 'DTN' }
+      let(:measurement_unit_qualifier_code) { 'Z' }
+
+      it { is_expected.to be_sucrose }
+    end
+
+    context 'when the unit is not an sucrose volume unit' do
```
alexdesi, line 136: when the unit is not a~n~ sucrose volume unit

spec/models/api/base_component_spec.rb
willfish: Discussed. We can but it would be a PITA/require updating all specs to be consistent

alexdesi (approved):
✅  🐒

PR #568, opened 2022-12-21
HOTT-2389 Add measure type 464

jebw:
### Jira link

HOTT-2389

### What?

I have added/removed/altered:

- [x] Added support for measure type 464

### Why?

I am doing this because:

- We wish to request these documents on the duty calculator

### Have you? (optional)

- Low


willfish:
So this is a bit of a weird measure in the sense that it's never going to be an additional duty row and its never going to be an option a trader can take.

Normally, when you want to prevent trade without a document you would add a measure condition to the relevant trade option measure (e.g. 103, 105, etc) that says without this document code you can't proceed.

Here, we have a measure that seems to globally apply to all of the options.

What I'd probably do is:

1. Rename the constant `Api::MeasureType::SUPPORTED_MEASURE_TYPE_IDS` to `Api::MeasureType::DOCUMENT_SUPPORTED_MEASURE_TYPE_IDS`
2. Merge in this new measure type id with a note indicating that this measure type is special because it is purely a stopping measure type and not one that yields duties
3. Add a small test to validate that the ApplicableDocumentCodeService will include this special measure type based on some input fixtures

I'd argue you don't really want a duty option/additional duty option at all in this unusual situation.

The functionality is working at the moment because we're essentially an additional duty option with no duties and so are bypassing adding the duties to the option table because of this guy, here: https://github.com/trade-tariff/trade-tariff-duty-calculator/blob/main/app/models/duty_calculator.rb#L30

willfish:
For reference: https://transformuk.atlassian.net/browse/HOTT-1455

willfish (approved):
Nice!

PR #619, opened 2023-04-04
HOTT-2989: Adds support for broken additional codes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2989

### What?

I have added/removed/altered:

- [x] Enable pluck on Base model
- [x] Removes unused method and spec
- [x] Adds support for broken additional codes

### Why?

I am doing this because:

- This is to make the duty calculator make more sense to users in the interim

### DIFF
--- a/app/models/api/commodity.rb
+++ b/app/models/api/commodity.rb
@@ -43,6 +43,8 @@ module Api
       formatted_code = "#{code[0..3]} #{code[4..5]} #{code[6..7]} #{code[8..9]}"
       return formatted_code if additional_code.blank?
 
+      additional_code = 'No additional code' if additional_code == 'none'
+
       "#{formatted_code} (#{additional_code})".html_safe
     end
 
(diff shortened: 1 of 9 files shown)


app/models/api/commodity.rb
```diff
@@ -91,22 +93,54 @@ def excise_measures
...
+      end
     end
 
     def additional_code_measures
       non_vat_import_measures.select do |measure|
-        measure_additional_code = measure.additional_code&.formatted_code
-        user_additional_code_answer = user_session.additional_code_for(measure.measure_type, source)
+        answer = user_session.additional_code_for(measure.measure_type.id)
+        is_additional_code_measure = measure.id.in?(additional_code_measure_sids)
+        has_answer = answer == if measure.additional_code.blank?
```
jebw, line 105: `measure.additional_code.presence&.code || 'none'` ?

jebw (approved):
LGTM

PR #620, opened 2023-04-06
HOTT-2989: Handle excise additional codes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2989

image

### What?

I have added/removed/altered:

- [x] Use a compatible multipart-post
- [x] Handle excise additional codes
- [x] Verify that excise additional codes are being picked

### Why?

I am doing this because:

- This was missed as part of the broken additional code work

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/trade-tariff/uktt.git
-  revision: 2f18bce22d5c69ab938e0b657f65db49acd3d25a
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -202,7 +202,7 @@ GEM
     mini_portile2 (2.8.1)
     minitest (5.18.0)
     msgpack (1.7.0)
-    multipart-post (2.3.0)
+    multipart-post (2.2.3)
```
willfish, line 205: This minor version change completely alters the require structure of the gem so breaks its faraday integration

PR #636, opened 2023-06-29
HOTT-3371 Use no-store to prevent http caching

jebw:
### Jira link

HOTT-3371

### What?

I have added/removed/altered:

- [x] Set cache-control to `no-store` 

### Why?

I am doing this because:

- It prevents any caching of the html content, since this is session dependant

- Low, should reduce risks of ending up in caches


willfish (approved):

willfish:
Don't have time to verify this right now so closing

PR #661, opened 2023-07-28
HOTT-3649: Adds evaluator for SPQ component

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3649

### What?

I have added/removed/altered:

- [x] Added SPQ component evaluation and specs

### Why?

I am doing this because:

- This is required so the new excise duties can be calculated properly


app/models/steps/excise.rb
```diff
@@ -19,10 +19,12 @@ def save
     def options
       available_additional_codes.map do |additional_code|
         code = additional_code['code'].sub('X', '')
+        overlay = additional_code['overlay'].to_s
+        overlay = "#{overlay} (X#{code})".html_safe
```
alexdesi, line 23: Just one assignment?
`overlay = "#{additional_code['overlay']} (X#{code})".html_safe`

spec/controllers/steps/confirmation_controller_spec.rb
```diff
@@ -44,16 +44,8 @@
...
-    end
-
-    it 'contains the links that allow users to go back and change their answers' do
-      expected_links.each do |link|
-        expect(response.body).to include(link)
-      end
-    end
-
+    it { expect(response.body).to include(expected_content) }
+    it { expected_links.each { |link| expect(response.body).to include(link) } }
```
alexdesi, line 48: nice cleaning!

alexdesi (approved):
💚

PR #691, opened 2023-09-18
Add guard for when the duty returns nil for a commodity

shjohnson:
### Jira link

HOTT-3916

### What?

I have added/removed/altered:

- [ ] Catch for when the duty returns nil instead of a numeric value

https://engine-le.sentry.io/issues/4438500131/?alert_rule_id=6424017&alert_type=issue&project=5701976&referrer=slack

### DIFF
--- a/app/models/api/base_component.rb
+++ b/app/models/api/base_component.rb
@@ -41,6 +41,12 @@ module Api
       spq: %w[SPQ],
     }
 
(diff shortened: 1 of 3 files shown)


shjohnson:
This wont get used as the current error message is adequate and this is a data problem not functionality issue.

PR #692, opened 2023-09-18
Change format of where we say not at risk to correct format

shjohnson:
### Jira link

HOTT-4020

### What?

I have added/removed/altered:

- [ ] Change `not `at risk` to `not at risk``


amberstarlight (approved):
Nice and simple. LGTM.

willfish (approved):

PR #693, opened 2023-09-18
Change reference to £500,000 to 2million where applicable

shjohnson:
### Jira link

HOTT-3938

### What?

I have added/removed/altered:

- [ ] Change reference to £500,000 to £2million

This goes first in the merge for the duty calculator work

### DIFF
--- a/app/controllers/steps/base_controller.rb
+++ b/app/controllers/steps/base_controller.rb
@@ -2,6 +2,7 @@ module Steps
   class BaseController < ::ApplicationController
     include CommodityHelper
     include ServiceHelper
+    include UkimsHelper
 
     rescue_from Errors::SessionIntegrityError, with: :handle_session_integrity_error
     rescue_from StandardError, with: :handle_exception
--- a/app/helpers/ukims_helper.rb
+++ b/app/helpers/ukims_helper.rb
@@ -1,0 +1,26 @@
+# Helper for UK Internal Market Scheme
+module UkimsHelper
+  CUT_OFF_DATE = Date.new(2023, 9, 30)
+
+  def uk_internal_market_scheme
+    if after_cut_off_date
+      'UK Internal Market Scheme'
+    else
+      'UK Trader Scheme'
+    end
+  end
+
+  def ukims_annual_turnover
+    if after_cut_off_date
+      '£2,000,000'
+    else
+      '£500,000'
+    end
+  end
+
+  private
+
+  def after_cut_off_date
+    user_session.import_date >= CUT_OFF_DATE
+  end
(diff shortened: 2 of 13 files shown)


app/controllers/steps/base_controller.rb
```diff
@@ -2,6 +2,7 @@ module Steps
   class BaseController < ::ApplicationController
     include CommodityHelper
     include ServiceHelper
+    include UkimsHelper
```
shjohnson, line 5: Might not need this later on but can stay for now i think...

app/helpers/ukims_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def uk_internal_market_scheme
+    if cut_off_validation
+      'UK Internal Market Scheme'
+    else
+      'UK Trader Scheme'
+    end
+  end
+
+  def ukims_annual_turnover
```
alexdesi, line 12: with _ looks better:
`uk_ims_annual_turnover`

app/helpers/ukims_helper.rb
```diff
@@ -0,0 +1,25 @@
+module UkimsHelper
```
alexdesi, line 1: Should we put a comment on what IMS stands for? (I did know what it was before this PR :) )

`# Helper for Internal Market Scheme`

app/helpers/ukims_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+  def uk_internal_market_scheme
+    if cut_off_validation
+      'UK Internal Market Scheme'
+    else
+      'UK Trader Scheme'
+    end
+  end
+
+  def ukims_annual_turnover
+    if cut_off_validation
```
alexdesi, line 13: maybe more explicit? 
validation sounds a bit generic

if `older_than_cut_off_date`
or
if 'more_recent_than_cut_off_date'

app/helpers/ukims_helper.rb
shjohnson: hmmm i kept it as all together as 'u-kims' is the way we are pronouncing uk-ims is confusing me ha..

app/helpers/ukims_helper.rb
shjohnson: Yes this might be a good idea at the top im happy to add a comment 👍 , I dont mind class level comments if not to long.

alexdesi (approved):
It looks good ✅ 
I am still not convinced about naming 
`ukims_annual_turnover`  and not `uk_ims_annual_turnover` but it's just a minor thing.

PR #695, opened 2023-09-19
Add copy for permitted processing (+or one subsequent entity)

shjohnson:
Includes: Replace UKTS and synonyms to UKIMS equivalents

### Jira link

HOTT-3939
HOTT-4020

### What?

I have added/removed/altered:

- [ ] Alter content to have alternative content in the bullet points for planned processing 
- [ ] Change wording for the 'UK Trader Scheme' to 'UK Internal Market Scheme'
- [ ]  Change links to the new links for the UKIMS application and info pages

### DIFF
--- a/app/controllers/steps/base_controller.rb
+++ b/app/controllers/steps/base_controller.rb
@@ -66,7 +66,7 @@ module Steps
     end
 
     def title
-      t("page_titles.#{@step.class.try(:id)}", uk_internal_market_scheme:, default: super)
+      t("page_titles.#{@step.class.try(:id)}", default: super)
     end
 
     def with_session_tracking
--- a/app/controllers/steps/base_controller.rb
+++ b/app/controllers/steps/base_controller.rb
@@ -66,7 +66,7 @@ module Steps
     end
 
     def title
-      t("page_titles.#{@step.class.try(:id)}", default: super)
+      t("page_titles.#{@step.class.try(:id)}", uk_internal_market_scheme:, default: super)
     end
 
     def with_session_tracking
--- a/app/helpers/ukims_helper.rb
+++ b/app/helpers/ukims_helper.rb
@@ -3,7 +3,7 @@ module UkimsHelper
   CUT_OFF_DATE = Date.new(2023, 9, 30)
 
   def uk_internal_market_scheme
-    if after_cut_off_date
+    if after_cut_off_date?
       'UK Internal Market Scheme'
     else
       'UK Trader Scheme'
@@ -11,7 +11,7 @@ module UkimsHelper
   end
 
   def ukims_annual_turnover
-    if after_cut_off_date
+    if after_cut_off_date?
 
       '£2,000,000'
     else
@@ -19,10 +19,35 @@ module UkimsHelper
     end
   end
 
+  def ukims_application_link
+    if after_cut_off_date?
+      'https://www.gov.uk/guidance/apply-for-authorisation-for-the-uk-internal-market-scheme-if-you-bring-goods-into-northern-ireland'
+    else
+      'https://www.gov.uk/guidance/apply-for-authorisation-for-the-uk-trader-scheme-if-you-bring-goods-into-northern-ireland-from-1-january-2021'
+    end
+  end
+
+  def ukims_info_url
+    if after_cut_off_date?
+      'https://www.gov.uk/government/news/windsor-framework-unveiled-to-fix-problems-of-the-northern-ireland-protocol'
+    else
+      'https://www.gov.uk/guidance/apply-for-authorisation-for-the-uk-trader-scheme-if-you-bring-goods-into-northern-ireland'
+    end
+  end
+
+  def explore_topic_title
+    if after_cut_off_date?
+      'Find out more about the Windsor Framework'
+    else
+      'Explore the topic'
+    end
+  end
+
   private
 
+  def after_cut_off_date?
+    return false unless user_session.import_date
 
-  def after_cut_off_date
     user_session.import_date >= CUT_OFF_DATE
   end
 end
--- a/app/helpers/ukims_helper.rb
+++ b/app/helpers/ukims_helper.rb
@@ -21,6 +21,7 @@ module UkimsHelper
 
   def ukims_application_link
     if after_cut_off_date?
+
       'https://www.gov.uk/guidance/apply-for-authorisation-for-the-uk-internal-market-scheme-if-you-bring-goods-into-northern-ireland'
     else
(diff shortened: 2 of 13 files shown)


app/helpers/ukims_helper.rb
```diff
@@ -41,6 +41,35 @@ def explore_topic_title
...
+            <li><strong>not for profit</strong> activities in Northern Ireland by the importer or one subsequent entity, where there is no subsequent sale of the processed goods</li>
+            <li>the final use of <strong>animal feed</strong> on premises located in Northern Ireland by the importer or one subsequent entity.</li>
+          </ul>
+                    PARAGRAPH
+
+                  else
+                    <<-PARAGRAPH
+          <ul>
+            <li>the sale of <strong>food</strong> to end consumers in the UK</li>
+            <li><strong>construction</strong>, where the processed goods form a permanent part of a structure that is constructed and located in NI by the importer</li>
```
alexdesi, line 61: The preview PARAGRAPH uses Northern Ireland instead of NI,
should we expand NI -> Northern Ireland as well here?

alexdesi (dismissed):
Nice one! ✅

shjohnson:
Im going to rebase this onto main for debugging reasons

app/controllers/steps/base_controller.rb
```diff
@@ -66,7 +66,7 @@ def initialize_commodity_context_service
     end
 
     def title
-      t("page_titles.#{@step.class.try(:id)}", default: super)
+      t("page_titles.#{@step.class.try(:id)}", uk_internal_market_scheme:, default: super)
```
willfish, line 69: I'm wondering about coupling the base controller with a concrete title setting. Is it better to just override the title in the underlying controller?

app/helpers/ukims_helper.rb
```diff
@@ -3,24 +3,79 @@ module UkimsHelper
...
+      'Find out more about the Windsor Framework'
+    else
+      'Explore the topic'
+    end
+  end
+
+  concerning :PlannedProcessingText do
+    def commercial_processing_hint_text
+      html_data = if after_cut_off_date?
+                    <<-PARAGRAPH
```
willfish, line 48: Wondering whether it would be cleaner to just chuck these into a conditionally selected partial?

shjohnson: I guess that is an option although this is included in the locales yml file and could get a bit messy either way. At least this way once a certain date has passed then we can just delete this code.

app/controllers/steps/base_controller.rb
shjohnson: Potentially although i feel it depends how many options are passed into the title tag maybe? Having these sorts of changes i think will always feel a little dirty and quick because of the purpose they are solving. I guess if we have a whole heap of variables being passed in for different steps and their titles then would make sense to move the title in that controller class.

app/helpers/ukims_helper.rb
willfish: This should sort out the indentation madness, too :tada:

app/helpers/ukims_helper.rb
willfish: We're not really coming up with the content and this has existed for a long time. I'd be tempted to not change it given all new trades will be in the other branch

app/helpers/ukims_helper.rb
```diff
@@ -3,24 +3,79 @@ module UkimsHelper
   CUT_OFF_DATE = Date.new(2023, 9, 30)
 
   def uk_internal_market_scheme
```
willfish, line 5: What do you think of a global renaming of this to `market_scheme_type`

app/controllers/steps/base_controller.rb
willfish: I'll have a look at the code a bit more and see if I'm missing something but this is basically breaching SOLID principles

willfish: Yeah, all we need to do to avoid mixing special titles is to update the trader scheme controller:

```ruby
module Steps
  class TraderSchemeController < BaseController
    helper_method :title

    def show
      @step = Steps::TraderScheme.new
    end

    def create
      @step = Steps::TraderScheme.new(permitted_params)

      validate(@step)
    end

    def title
      t("page_titles.#{@step.class.try(:id)}", uk_internal_market_scheme:, default: super)
    end

    private

    def permitted_params
      params.require(:steps_trader_scheme).permit(
        :trader_scheme,
      )
    end
  end
end
```

app/helpers/ukims_helper.rb
willfish: But we have an idiom/way of doing things and even if you delete the code you'd be diverging from that just for this specific case?

app/helpers/ukims_helper.rb
willfish: I think given we're going to delete the old one in a year or so we can ignore this

shjohnson: Until another scheme comes out 😂 , i think this could still be worth it if im making changes elsewhere

app/helpers/ukims_helper.rb
shjohnson: Yeah il have a look at this, that does look better.

shjohnson: The indentation was kind of on purpose but rubocop also didnt help lol

willfish: [shakes fist at rubocop]

app/views/steps/planned_processing/show.html.erb
```diff
@@ -9,7 +9,7 @@
     <div class='govuk-!-margin-bottom-3' >
       <%= f.govuk_radio_button :planned_processing, 'without_any_processing', label: { text: t('planned_processing.without_any_processing.label_text_html') } %>
     </div>
-    <%= f.govuk_radio_button :planned_processing, 'commercial_processing', label: { text: t('planned_processing.commercial_processing.label_text_html') }, hint: { text: t('planned_processing.commercial_processing.hint_text_html') } %>
+    <%= f.govuk_radio_button :planned_processing, 'commercial_processing', label: { text: t('planned_processing.commercial_processing.label_text_html') }, hint: { text: commercial_processing_hint_text, commercial_processing_hint_text: commercial_processing_hint_text } %>
```
willfish, line 12: Do you need the second hint key/value, here?

willfish (approved):
LGTM

app/views/steps/planned_processing/show.html.erb
shjohnson: Sorry what do you mean? there was a formatting issue which brakeman picked up and i just pushed.

PR #727, opened 2023-11-29
BAU: Update base domain

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Update base domain for production

### Why?

I am doing this because:

- This is currently incorrect/pointing to the old test domain


amberstarlight (approved):

willfish:
Fixing dev slowness in a bit

PR #730, opened 2023-12-07
HOTT-4467: Add CSP policy to Duty Calculator app

rasikasri:
### Jira link

[HOTT-4467](https://transformuk.atlassian.net/browse/HOTT-4467)

### What?

I have added/removed/altered:

- [ ] Added content security policy

### Why?

I am doing this because:
We should add a content security policy to reduce the risk of XSS problems in the sites

### Have you? (optional)

- Makes changes to our complex routing setup that may affect apis to proxying to backend

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-nodejs 16.5.0
+nodejs 20.10.0
 yarn 1.22.10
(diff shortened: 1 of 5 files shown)


jebw (approved):
LGTM - only query is the change to the 404 response, could probably represent as text/plain rather than text/html but seems minor

jebw (approved):
LGTM

PR #737, opened 2023-12-12
Update rails 7.1.2

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-4624

### What?
Update Rails from 7.0 to 7.1.2


config/environments/development.rb
```diff
@@ -6,7 +6,7 @@
   # In the development environment your application's code is reloaded any time
   # it changes. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = false
+  config.enable_reloading = true
```
alexdesi, line 9: NOTE: Even though raise_on_missing_translations was set to true in Rails 7.0, it did not raise exceptions in cases of missing translations.

config/environments/test.rb
```diff
@@ -43,8 +44,11 @@
   config.active_support.disallowed_deprecation_warnings = []
 
   # Raises error for missing translations.
-  config.i18n.raise_on_missing_translations = true
+  config.i18n.raise_on_missing_translations = false
```
alexdesi, line 47: NOTE: Even though raise_on_missing_translations was set to true in Rails 7.0, it did not raise exceptions in cases of missing translations.

Gemfile
```diff
@@ -2,19 +2,15 @@ source 'https://rubygems.org'
 
 ruby File.read('.ruby-version').chomp
 
-gem 'actionpack', '~> 7'
-gem 'actionview', '~> 7'
-gem 'activemodel', '~> 7'
-gem 'activerecord', '~> 7'
-gem 'activesupport', '~> 7'
+gem 'rails', '~> 7.1'
```
jebw, line 5: This will pull in actiontext + activestorage + activerecord - I don't think that's necessary for an api only app

Can you swap back to the old lines but update to 7.1

config/environments/production.rb
```diff
@@ -31,35 +30,43 @@
...
 
-  # Include generic and useful information about system operation, but avoid logging too much
-  # information to avoid inadvertent exposure of personally identifiable information (PII).
-  config.log_level = :info
+  # Assume all access to the app is happening through a SSL-terminating reverse proxy.
+  # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
+  # config.assume_ssl = true
+
+  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
+  config.force_ssl = true
```
jebw, line 38: @alexdesi This is the same as frontend? If you're using this option, you'll need to enable `assume_ssl` I think ? 

@willfish @amberstarlight I'm not familiar on why this wasn't working before but it does appear to be working without `assume_ssl`????

config/environments/test.rb
jebw: This is only in the test environment but it should be left on, missing translation _should_ fail the spec

config/initializers/content_security_policy.rb
```diff
@@ -1,9 +1,3 @@
-# Be sure to restart your server when you modify this file.
```
jebw, line 1: Worth leaving this note in place - might help future developers confused by their changes are having no impact - below is the rails 7.1 content which should be here

```
# Be sure to restart your server when you modify this file.

# Define an application-wide content security policy.
# See the Securing Rails Applications Guide for more information:
# https://guides.rubyonrails.org/security.html#content-security-policy-header
```

jebw (changes requested):
Looks good - couple of minor points to address and I think you need clarity on why `force_ssl` got turned off originally

config/environments/test.rb
alexdesi: I agree with that,  but,
consider that it seems not to be working on the main branch (see my note):
in the spec:
spec/models/row_to_ni_duty_calculator_spec.rb
there are many missing translations, and it does not trigger errors on main :man_shrugging: 

However, it does trigger errors after the update (correct behaviour)
 
I guess the right path is to set it to true and fix the broken specs, which are not broken on Main because it does not consider that flag.

Gemfile
alexdesi: I had some trouble with that, some dependencies had conflicts.
I'll try it again, sometimes changing the order fix the them.

config/environments/production.rb
willfish: CDN and load balancer terminate SSL and reverse proxy onto the apps, yeah.

The deployment will fail with force_ssl = true and we should instead be using assume_ssl = true

https://github.com/trade-tariff/trade-tariff-platform-aws-terraform/blob/main/environments/common/application-load-balancer/alb.tf#L26-L36

The load balancer checks the health of the apps via these endpoints over http:

```terraform
locals {
  services = {
    admin = {
      host             = ["admin.*"]
      healthcheck_path = "/healthcheckz"
    }

    signon = {
      host             = ["signon.*"]
      healthcheck_path = "/healthcheck/live"
    }

    backend_uk = {
      paths            = ["/uk/api/beta/*"]
      healthcheck_path = "/healthcheckz"
    }

    backend_xi = {
      paths            = ["/xi/api/beta/*"]
      healthcheck_path = "/healthcheckz"
    }

    duty_calculator = {
      paths            = ["/duty-calculator/*"]
      healthcheck_path = "/healthcheckz"
    }

    search_query_parser = {
      paths            = ["/api/search/*"]
      healthcheck_path = "/healthcheckz"
    }

    frontend = {
      paths            = ["/*"]
      healthcheck_path = "/healthcheckz"
    }
  }
}
```

willfish: https://github.com/rails/rails/pull/47139/commits/17ee01f22b1a59987c47e78625b8f8fd9513a740

app/models/row_to_ni_duty_calculator.rb
```diff
@@ -31,7 +31,10 @@ def options
 
                  uk_only = cheapest_xi_option.nil?
 
-                 footnote_suffix = I18n.t("row_to_ni_measure_type_footnotes_suffixes.#{category}.uk_only.#{option.source}_html", uk_only_text: uk_only_text_for(category, option, uk_option)) if uk_only
+                 footnote_suffix = if uk_only
+                                     I18n.t("row_to_ni_measure_type_footnotes_suffixes.#{category}.uk_only.#{option.source}_html",
+                                            uk_only_text: uk_only_text_for(category, option, uk_option))
+                                   end
```
alexdesi, line 37: This is just a tiny refactor to make the row shorter and readable.

config/environments/production.rb
alexdesi: I've also set
`config.assume_ssl = true`
for the TT Admin

willfish (approved):

PR #738, opened 2023-12-14
BAU: Fix image-name in prod release job

amberstarlight:
### What?

I have added/removed/altered:

- Fixed broken `image-name` parameter in release job

### Why?

I am doing this because:

- It's wrong and blocking releases


alexdesi (approved):
Nice one! 
thanks :white_check_mark:

PR #739, opened 2023-12-15
HOTT-4630: Extend DutyCalculator ErrorsController

rasikasri:
### Jira link

[HOTT-4630](https://transformuk.atlassian.net/browse/HOTT-4630)

### What?
- added custom handling for various HTTP status
- removed old error_controller_spec.rb and added a new request/error_controller_spec.rb
- refactor various error views into an unique template for error `error.html.erb`  

### Why?
- To give better error messages/experience to the users
- To be consistent with all the other TT projects


spec/requests/errors_controller_spec.rb
```diff
@@ -0,0 +1,96 @@
+RSpec.describe ErrorsController, type: :request do
+  include ApplicationHelper
+
+  subject(:rendered) { make_request && response }
+
+  let(:json_response) { JSON.parse(rendered.body) }
```
shjohnson, line 6: I always think it would be nice to move this type of method in a global helper but then depends what we are passing into the body part

shjohnson: Can pass body here i guess as use it below

config/routes.rb
```diff
@@ -64,7 +64,11 @@
     get 'prefill', to: 'steps/prefill_user_session#show'
   end
 
+  get '400', to: 'errors#bad_request', via: :all
```
shjohnson, line 67: 👏

app/views/errors/error.html.erb
```diff
@@ -0,0 +1,12 @@
+<main role="main" class="govuk-main-wrapper" id="main-content">
```
shjohnson, line 1: Have you got a screenshot of what this looks like for the PR purposes?

app/controllers/errors_controller.rb
```diff
@@ -3,25 +3,90 @@ class ErrorsController < ApplicationController
...
-      format.all { render status: :not_found, body: nil }
+      format.all { render status: :not_found, plain: 'Resource not found'  }
     end
   end
 
-  def internal_server_error
+  # /405
+  def method_not_allowed
+    message = "We're sorry, but this request method is not supported.<br>
+                Please contact support for assistance or try a different request.".html_safe
```
shjohnson, line 32: Should we add a support email here?

shjohnson: Probably. general information for the other times we suggest to contact support

shjohnson (approved):
Couple of comments but nothing blocking LGTM

willfish (approved):

PR #745, opened 2024-01-05
HOTT-4446: Adds HEALTHCHECK

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4446

### What?

ECS healthchecks are managed separately:

https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_HealthCheck.html

I have added/removed/altered:

- [x] Added HEALTHCHECK to Dockerfile
- [x] Make Dockerfile run locally with minimum-required configuration

### Why?

I am doing this because:

- This is remediation work from our pentest report


config/environments/production.rb
```diff
@@ -32,7 +32,7 @@
 
   # Assume all access to the app is happening through a SSL-terminating reverse proxy.
   # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
-  config.assume_ssl = true
+  config.assume_ssl = ENV.fetch('RAILS_ASSUME_SSL', "true") == "true"
```
willfish, line 35: We don't want to assume reverse proxy that terminates SSL when running locally

amberstarlight (approved):

PR #753, opened 2024-01-18
BAU: Fix issue with deployment

amberstarlight:
### What?

I have added/removed/altered:

- Added `GOVUK_APP_DOMAIN` env var (this is necessary for a gem we're not using).
- Removed NewRelic things that aren't useful.

### Why?

I am doing this because:

- Deploys are failing silently because of this error


jebw (dismissed):
LGTM

jebw (approved):

PR #764, opened 2024-03-12
HOTT-5136: Added top and bottom choice selection banners to Duty calc

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-5136

### What?

I have added/removed/altered:

- [ ] Added top and bottom choice selection banners to Duty calc

### Why?

I am doing this because:

-  We need to display feedback banner's on duty calc so we can collect user feedback

Screenshot 2024-03-12 at 11 03 19

Screenshot 2024-03-12 at 11 04 12

### DIFF
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -48,8 +48,12 @@ module ServiceHelper
     service_url_for("/commodities/#{commodity_code}")
   end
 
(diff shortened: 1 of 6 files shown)


HWallenberg (approved):
lgtm :+1:

PR #815, opened 2024-09-17
OTT-275: Adds handling for frontend-forwarded import date

willfish:
### Jira link

OTT-275

### What?

I have added/removed/altered:

- [x] Added handling for custom as_of dates from the frontend

### Why?

I am doing this because:

- This enables us to let users use the duty calculator even when the
commodity is not current (and therefore visible via the api on today's
date).


HWallenberg (approved):
lgtm 🚀

PR #823, opened 2025-01-09
HMRC-538 Fix GTM Tags

neilmiddleton:
### Jira link

HMRC-538

### What?

I have updated Google Tag Manager code with the correct Container IDs to match the commodities pages.


HWallenberg:
@neilmiddleton should we parameterise tags per env like the main site?

Tags for each env documented internally [here:](https://transformuk.atlassian.net/wiki/spaces/HOTT/pages/21605089281/3.6.9+-+Google+Tag+Manager+GTM+Configuration )

Jira: https://transformuk.atlassian.net/browse/GL-538.  

PR: https://github.com/trade-tariff/trade-tariff-frontend/pull/2014

No problem if not.

neilmiddleton:
@HWallenberg I know what you mean, but I think that's bringing in a load of complexity which we probably wouldn't get much benefit on given this is such as low traffic app outside production.

I do wonder about `usage_enabled?` being restricted to production only though?

HWallenberg (approved):

PR #825, opened 2025-01-22
HMRC-583 Remove Robots meta tag

neilmiddleton:
### Jira link

HMRC-583

### What?

Remove robots meta tag - allow robots.txt to work it's magic.


HWallenberg (approved):
i like magic 🪄

PR #826, opened 2025-01-28
HMRC-580-Update-GTM-Implementation

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HMRC-580

### What?

- [x] I have refactored GTM to be cookie preference sensitive and parameterised GTM tags as per environment

### Why?

I am doing this because:

- The TradeTariffFrontEnd already does this and this behaviour is to be replicated in DutyCalculator


amberstarlight (commented):
Terraform changes look good to me - the ID isn't a secret so fine to be put here 😄

app/webpacker/src/javascripts/cookie-manager.js
```diff
@@ -0,0 +1,127 @@
...
+        return true;
+      } else if (rememberSettings === 'false') {
+        return false;
+      } else {
+        return rememberSettings;
+      }
+    } else {
+      return false;
+    }
+  }
```
EraliaKats, line 26: QQ - I see this pattern used extensively in your code. I'm trying to understand what other values rememberSettings could take other than boolean? It's a tad bit difficult to read imo

Personally, for easy true or false logic I'm more of a fan of the `?` operator, makes it a lil less complicating

app/webpacker/src/javascripts/cookie-manager.js
```diff
@@ -0,0 +1,106 @@
...
+
+  showAcceptRejectCookiesBanner() {
+    if (!this.getCookiesPolicy()) {
+      return true;
+    } else {
+      return false;
+    }
+  }
```
EraliaKats, line 43: would you mind refactoring this 👉 👈

app/webpacker/src/javascripts/cookie-manager.js
```diff
@@ -0,0 +1,106 @@
...
+
+  showHideConfirmCookiesBanner() {
+    if (this.getCookiesHideConfirm() === null) {
+      return true;
+    } else {
+      return false;
+    }
```
EraliaKats, line 50: would you mind refactoring this as well 👉 👈

app/webpacker/src/javascripts/cookie-manager.js
```diff
@@ -0,0 +1,106 @@
...
+    if (this.getCookiesTreeOpenClosedDefault()) {
+      const value = this.getCookiesTreeOpenClosedDefault();
+
+      if (value === 'open') {
+        return true;
+      } else {
+        return false;
+      }
+    } else {
+      return true;
```
EraliaKats, line 33: looks like we could use the same pattern here

PR #864, opened 2025-03-26
HMRC-909: Enable CD in the production workflow

willfish:
### Jira link

[HMRC-909](https://transformuk.atlassian.net/browse/HMRC-909)

### What?

I have added/removed/altered:

- [x] Added CD for the production workflow

### Why?

I am doing this because:

- We now have e2e tests and these cover a full duty calculation


HWallenberg (approved):
🚀

PR #902, opened 2025-06-03
Scale down DC to zero

neilmiddleton:
### Jira link

HMRC-694

### What?

Scale down DC to zero whilst we let the migrated version in FE soak for a bit

Depends on https://github.com/trade-tariff/trade-tariff-frontend/pull/2339

### DIFF
--- a/terraform/config_production.tfvars
+++ b/terraform/config_production.tfvars
@@ -3,5 +3,5 @@ environment   = "production"
 cpu           = 1024
 memory        = 2048
 service_count = 3
-min_capacity  = 2
-max_capacity  = 5
+min_capacity  = 0
+max_capacity  = 0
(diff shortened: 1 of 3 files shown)


terraform/config_production.tfvars
```diff
@@ -3,5 +3,5 @@ environment   = "production"
 cpu           = 1024
 memory        = 2048
 service_count = 3
-min_capacity  = 2
```
willfish, line 6: Aren't people still using the production one whilst we bed in with the merged dev/staging versions?

neilmiddleton: If it's working in staging it'll work in production most likely.   If we're confident enough to merge we should be confident enough to run in prod.

willfish (approved):

